PR #2, opened 2016-05-06
Add per-retry request logging hook

jen20:


client_test.go
```diff
@@ -188,6 +200,38 @@ func TestClient_Get(t *testing.T) {
 	resp.Body.Close()
 }
 
+func TestClient_RequestLogHook(t *testing.T) {
+	// Mock server which always responds 500.
```
ryanuber: But does it? :) Looks like this comment is just outdated.

client_test.go
```diff
@@ -188,6 +200,38 @@ func TestClient_Get(t *testing.T) {
...
+		if r.RequestURI != "/foo/bar" {
+			t.Fatalf("bad uri: %s", r.RequestURI)
+		}
+		w.WriteHeader(200)
+	}))
+	defer ts.Close()
+
+	retries := -1
+
+	client := NewClient()
```
ryanuber: Can we set a logger on the new client, write to it in the callback and check its buffer? Or at least just ensure the same logger instance is passed through.

ryanuber:
@jen20 just minor comments but LGTM!

ryanuber:
We should probably also test the callback with a failing HTTP request to ensure we get a callback for each retry.

jen20:
Thanks for the review @ryanuber - all issues have been addressed. Another instance of the out-of-date comment is also removed in this.

ryanuber:
LGTM! Merge away :ship:

PR #4, opened 2016-08-05
Add Client.CheckRetry callback

jbardin:
There currently is no way to set a policy to determine what
constitutes a retry condition. We can't stop retries on an error we know
is unrecoverable, and we can't continue retrying on an error that we
know is temporary. For example, we probably want to backoff and retry on
429, because that's what it's for, though some systems may send a
different code and/or message, and may even provide a retry time.

This adds a CheckRetry callback to the client that will be called after
each request. If CheckRetry returns an error, the client will
immediately return the response and that error (which may just be the
error returned from the original request).

### DIFF
--- a/client.go
+++ b/client.go
@@ -39,9 +39,9 @@ var (
 	// a new client. It is purposely private to avoid modifications.
 	defaultClient = NewClient()
 
-	// We need to consume response bodied to maintain http connections, but
+	// We need to consume response bodies to maintain http connections, but
 	// limit the size we consume to respReadLimit.
-	respReadLimit = int64(4 << 10)
+	respReadLimit = int64(4096)
 )
 
 // LenReader is an interface implemented by many in-memory io.Reader's. Used
@@ -114,6 +114,15 @@ type Client struct {
 	// ResponseLogHook allows a user-supplied function to be called
 	// with the response from each HTTP request executed.
 	ResponseLogHook ResponseLogHook
+
+	// CheckRetry specifies a policy for handling retries. It is called
+	// following each request with the response and error values returned by
+	// the http.Client. If CheckRetry returns false, the Client stops retrying
+	// and returns the response to the caller. If CheckRetry returns an error,
+	// that error value is returned in lieu of the error from the request. If
+	// an error is returned, it is up to the CheckResponse callback to properly
+	// close any response body before returning.
+	CheckRetry func(resp *http.Response, err error) (bool, error)
 }
 
 // NewClient creates a new Client with default settings.
@@ -127,9 +136,30 @@ func NewClient() *Client {
 	}
 }
 
+// DefaultRetryPolicy provides a default callback for Client.CheckRetry, which
+// will retry on connection errors and server errors.
+func DefaultRetryPolicy(resp *http.Response, err error) (bool, error) {
+	if err != nil {
+		return true, err
+	}
+	// Check the response code. We retry on 500-range responses to allow
+	// the server time to recover, as 500's are typically not permanent
+	// errors and may relate to outages on the server side. This will catch
+	// invalid response codes as well, like 0 and 999.
+	if resp.StatusCode == 0 || resp.StatusCode >= 500 {
+		return true, nil
+	}
+
+	return false, nil
+}
+
 // Do wraps calling an HTTP method with retries.
 func (c *Client) Do(req *Request) (*http.Response, error) {
 	c.Logger.Printf("[DEBUG] %s %s", req.Method, req.URL)
+
+	if c.CheckRetry == nil {
+		c.CheckRetry = DefaultRetryPolicy
+	}
 
 	for i := 0; ; i++ {
 		var code int // HTTP response code
@@ -147,27 +177,34 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 
 		// Attempt the request
 		resp, err := c.HTTPClient.Do(req.Request)
+
+		// Check if we should continue with retries.
+		checkOK, checkErr := c.CheckRetry(resp, err)
+
 		if err != nil {
 			c.Logger.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
-			goto RETRY
+		} else {
+			// Call this here to maintain the behavior of logging all requests,
+			// even if CheckRetry signals to stop.
+			if c.ResponseLogHook != nil {
+				// Call the response logger function if provided.
+				c.ResponseLogHook(c.Logger, resp)
+			}
 		}
-		code = resp.StatusCode
 
-		// Call the response logger function if provided.
-		if c.ResponseLogHook != nil {
-			c.ResponseLogHook(c.Logger, resp)
+		// Now decide if we should continue.
+		if !checkOK {
+			if checkErr != nil {
+				err = checkErr
+			}
+			return resp, err
 		}
 
-		// Check the response code. We retry on 500-range responses to allow
-		// the server time to recover, as 500's are typically not permanent
-		// errors and may relate to outages on the server side.
-		if code%500 < 100 {
+		// We're going to retry, consume any response to reuse the connection.
+		if err == nil {
 			c.drainBody(resp.Body)
-			goto RETRY
 		}
-		return resp, nil
 
-	RETRY:
 		remain := c.RetryMax - i
 		if remain == 0 {
 			break
@@ -191,6 +228,10 @@ func (c *Client) drainBody(body io.ReadCloser) {
 	defer body.Close()
 	_, err := io.Copy(ioutil.Discard, &io.LimitedReader{R: body, N: respReadLimit})
 	if err != nil {
+		// don't log this if the body was closed by the ResponseLogHook or CheckRetry
+		if strings.Contains(err.Error(), "read on closed response body") {
+			return
+		}
 		c.Logger.Printf("[ERR] error reading response body: %v", err)
 	}
 }
--- a/client.go
+++ b/client.go
@@ -38,6 +38,10 @@ var (
 	// defaultClient is used for performing requests without explicitly making
 	// a new client. It is purposely private to avoid modifications.
 	defaultClient = NewClient()
+
+	// We need to consume response bodied to maintain http connections, but
+	// limit the size we consume to respReadLimit.
+	respReadLimit = int64(4 << 10)
 )
 
 // LenReader is an interface implemented by many in-memory io.Reader's. Used
@@ -158,7 +162,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		// the server time to recover, as 500's are typically not permanent
 		// errors and may relate to outages on the server side.
 		if code%500 < 100 {
-			resp.Body.Close()
+			c.drainBody(resp.Body)
 			goto RETRY
 		}
 		return resp, nil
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -38,6 +39,10 @@ var (
 	// defaultClient is used for performing requests without explicitly making
 	// a new client. It is purposely private to avoid modifications.
 	defaultClient = NewClient()
+
+	// We need to consume response bodied to maintain http connections, but
```
ryanuber: Misspelled "bodies" (?). I also wonder if it makes sense to limit this, as it makes it a bit unpredictable what the client will actually do or when it will re-use a connection. Thoughts?

client.go
```diff
@@ -38,6 +39,10 @@ var (
...
+	// limit the size we consume to respReadLimit.
+	respReadLimit = int64(4 << 10)
```
ryanuber: I would probably just specify 4096 here for clarity, but a comment in-line would also suffice.

client.go
```diff
@@ -123,10 +136,32 @@ func NewClient() *Client {
 	}
 }
 
+// ErrStopRetry is returned by a CheckRetry callback to stop retries, but
+// still have the client return the original error.
+var ErrStopRetry = errors.New("stop retry")
```
ryanuber: Instead of having a magic error, I'd rather have the `CheckRetry` return a `bool, error` tuple, the bool indicating if we should stop or not.

client.go
```diff
@@ -123,10 +136,32 @@ func NewClient() *Client {
...
+
 // Do wraps calling an HTTP method with retries.
 func (c *Client) Do(req *Request) (*http.Response, error) {
 	c.Logger.Printf("[DEBUG] %s %s", req.Method, req.URL)
 
+	if c.CheckRetry == nil {
+		// Check the response code. We retry on 500-range responses to allow
```
ryanuber: Looks like this comment is duplicated now. Let's reword and squash the two near the spot that actually checks the response code.

client.go
```diff
@@ -123,10 +136,32 @@ func NewClient() *Client {
...
+		// the server time to recover, as 500's are typically not permanent
+		// errors and may relate to outages on the server side.
+		c.CheckRetry = func(resp *http.Response, err error) error {
```
ryanuber: Let's make this a package-level function and default to it rather than inlining. It would also make sense to do this default assignment in `NewClient()` rather than the request path.

client.go
```diff
@@ -123,10 +136,32 @@ func NewClient() *Client {
...
+			if err != nil {
+				return nil
+			}
+			// continue with our default policy of retrying server errors
+			// also checking for the occasional odd response codes.
+			if resp.StatusCode == 0 || resp.StatusCode >= 500 {
+				return nil
+			}
+
+			return ErrStopRetry
```
ryanuber: Seems we should close out the body here since that is expected of the `CheckRetry` func.

client.go
```diff
@@ -143,27 +178,34 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
...
 		resp, err := c.HTTPClient.Do(req.Request)
+
+		// check if we should continue with retries
+		checkErr := c.CheckRetry(resp, err)
+
 		if err != nil {
 			c.Logger.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
-			goto RETRY
+		} else {
+			// call this here to maintain the behavior of logging all requests,
```
ryanuber: This will get skipped on all non-nil errors right? including the `ErrStopRetry`. I think we need to move this out.

client.go
```diff
@@ -182,6 +224,19 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		req.Method, req.URL, c.RetryMax+1)
 }
 
+// Try to read the response body so we can reuse this connection.
+func (c *Client) drainBody(body io.ReadCloser) {
+	defer body.Close()
```
ryanuber: I am a bit confused as to whether the `CheckRetry` is supposed to close the response body or not. Can you clarify?

ryanuber:
@jbardin this makes sense to me! Left some feedback. My only super nit-picky thing is to capitalize/punctuate the comments.

client.go
jbardin: It's just a trade-off for the time to drain the connection vs the time to setup a new one. The client should appear to behave the same either way, but limiting the reader prevent us from waiting for some very large, unexpected response when we could just as well drop the connection.

client.go
jbardin: SGTM, I'll refactor this to match.

client.go
jbardin: Thanks, it was here because it required a closure in an earlier commit to pass all tests. I can promote it now.

client.go
jbardin: We're only required to close the body if we return our own error when there was a valid response. Since this just signals to stop retrying, the caller can handle the `resp, error` combination normally.  (this will still apply with the new `(bool, error)` signature too)

client.go
jbardin: `ErrStopRetry` is check seprately, this will be more clear with the refactor

jbardin: This still looks pretty much the same. We guard this off in the else clause here, because we can't call `ResponseLogHook` if there was an error.  I agree this is little awkward, but it ensures the existing behavior was maintained where the logger is called even when we don't retry. I would normally have just dropped this down into the retry section where we drain the body.

jbardin:
@ryanuber: Refactored CheckRetry to use `(bool, error)`.
let me know if the refactoring and new comments clarify this for you, or if I missed anything else.

client.go
ryanuber: Makes sense. :+1:

client.go
```diff
@@ -110,6 +114,15 @@ type Client struct {
...
 	ResponseLogHook ResponseLogHook
+
+	// CheckRetry specifies a policy for handling retries. It is called
+	// following each request with the response and error values returned by
+	// the http.Client. If CheckRetry returns false, the Client stops retrying
+	// and returns the response to the caller. If CheckRetry returns an error,
+	// that error value is returned in lieu of the error from the request. If
+	// an error is returned, it is up to the CheckResponse callback to properly
+	// close any response body before returning.
+	CheckRetry func(resp *http.Response, err error) (bool, error)
```
ryanuber: Can we make this an interface similar to RequestLogHook and ResponseLogHook? It would be nice to have as part of the public API for any cases where you need to return one of these from some other function. That would also expose the description in the godoc so that the meaning of the `(bool, error)` is clear at a glance.

client.go
```diff
@@ -123,10 +136,31 @@ func NewClient() *Client {
...
+	}
+
+	return false, nil
+}
+
 // Do wraps calling an HTTP method with retries.
 func (c *Client) Do(req *Request) (*http.Response, error) {
 	c.Logger.Printf("[DEBUG] %s %s", req.Method, req.URL)
 
+	if c.CheckRetry == nil {
```
ryanuber: Can you move this into the `NewClient()`? We handle other default initialization there, like the default HTTP client, and I think this probably belongs there too.

client.go
```diff
@@ -182,6 +223,19 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		req.Method, req.URL, c.RetryMax+1)
 }
 
+// Try to read the response body so we can reuse this connection.
+func (c *Client) drainBody(body io.ReadCloser) {
+	defer body.Close()
+	_, err := io.Copy(ioutil.Discard, &io.LimitedReader{R: body, N: respReadLimit})
+	if err != nil {
+		// don't log this if the body was closed by the ResponseLogHook or CheckRetry
+		if strings.Contains(err.Error(), "read on closed response body") {
```
ryanuber: Can we just close the response body automatically near line 198 instead, and update the documentation such that the body close is handled for the client? In that case we know we are not retrying and that a custom error was returned. I always like to avoid switching on error strings when possible.

client_test.go
```diff
@@ -318,6 +319,71 @@ func TestClient_ResponseLogHook(t *testing.T) {
...
+		}
+
+		return false, retryErr
+	}
+
+	// CheckRetry should return our retryErr value and stop the retry loop.
+	_, err := client.Get(ts.URL)
+
+	if called != 1 {
+		t.Fatal("we didn't retry the correct number of times")
```
ryanuber: I would `Fatalf` this and show the number of tries.

client_test.go
```diff
@@ -318,6 +319,71 @@ func TestClient_ResponseLogHook(t *testing.T) {
...
+	}
+
+	if err == nil {
+		t.Fatal("expected error from request")
+	}
+
+	if err != retryErr {
```
ryanuber: This also covers the nil error case.

client_test.go
```diff
@@ -318,6 +319,71 @@ func TestClient_ResponseLogHook(t *testing.T) {
...
+		return false, nil
+	}
+
+	_, err := client.Get(ts.URL)
+
+	if called == 0 {
+		t.Fatal("CheckRetry wasn't called")
+	}
+
+	if called > 1 {
```
ryanuber: Let's just combine these to `called != 1`, and say like `expected 1 call to CheckRetry, got %d`

ryanuber:
@jbardin just a few more minor things, but this looks great! Thanks for tackling this. 💪

client.go
jbardin: I'm not sure I understand, because this will be exposed in the godoc since both Client and CheckRetry are exported. Function type struct fields are standard for callbacks, see Client.CheckRedirect, Transport.Dial, DialTLS, Proxy, Server.ConnState in the http package for example. And I'm not sure what an interface would provide over the ability to use a function literal and closure.

client.go
jbardin: We can add it there too, but calling a default when there's a nil value is pretty standard form for exported types to prevent panics and make the exported type usable without knowing the full initialization steps. This is just going by precedent, like tls.Config, http.Transport, etc.

client.go
jbardin: Yeah, I didn't like this either. I think it was required for some tests on an earlier iteration, but is no longer needed. I'll just pull this out and make a note in the docs.

jbardin:
Pushed a fix for the small stuff, and responded about the CheckRetry interface and initialization. I'm just trying to match the behavior of preexisting packages, but I'll leave the final call up to you.

client.go
ryanuber: Fair point about the http package. I think we should be consistent though, and since this is an existing public package, we should probably keep the interfaces. Take a look at the RequestLogHook and ResponseLogHook, they are just function interfaces so function literals can still be used in the same way. I am in the habit of using an interface when multiple implementations are expected.

client.go
ryanuber: This is a good point. How about we move the default HTTP client setup into `Do()` then as well, and just document behavior of passing nil for either?

client.go
jbardin: Oh, I have to admit I thought you meant they were actual interface types, and required methods to implement. The function types aren't a big deal. I'll change up the code to match your existing pattern.

client.go
jbardin: Yes, the split in functionality isn't good, and we should setup everything in one place. I like checking the fields in `Do()`, but there's no good way to check for unset values for `RetryWaitMin` and `RetryMax`which could have been set to 0 by the caller without ill effect.

So now that I think about it, since `Logger` will panic if it's not set currently, I'll move the CheckRetry into the NewClient function too.

jbardin:
Made a CheckRetry type, and put all initialization in the NewClient func.

ryanuber:
@jbardin LGTM!

PR #8, opened 2017-04-21
Don't retry on URL scheme and other non-retryable errors

phinze:
In scenarios with bad URLs like "not/a/uri", the default retry policy
was retrying even though the problem would never go away.

Turns out net.Error and url.Error have a `Temporary() bool` interface
they used to indicate whether or not they are retryable.

By checking for and respecting this interface, we can skip retries when
they are not going to make any difference.

### DIFF
--- a/client.go
+++ b/client.go
@@ -142,11 +142,23 @@ func NewClient() *Client {
 	}
 }
 
+type temporary interface {
+	Temporary() bool
+}
+
 // DefaultRetryPolicy provides a default callback for Client.CheckRetry, which
 // will retry on connection errors and server errors.
 func DefaultRetryPolicy(resp *http.Response, err error) (bool, error) {
 	if err != nil {
-		return true, err
+		retryable := true
+		switch t := err.(type) {
+		// net.Error and url.Error have a Temporary() bool that tells us if their
+		// errors are retryable, rather than typecheck for the specific error
+		// classes, we just use that same interface
+		case temporary:
+			retryable = t.Temporary()
+		}
+		return retryable, err
 	}
 	// Check the response code. We retry on 500-range responses to allow
 	// the server time to recover, as 500's are typically not permanent
@@ -184,7 +196,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		checkOK, checkErr := c.CheckRetry(resp, err)
 
 		if err != nil {
-			c.Logger.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
+			c.Logger.Printf("[ERR] %s %s request failed: %v (%T)", req.Method, req.URL, err, err)
 		} else {
 			// Call this here to maintain the behavior of logging all requests,
 			// even if CheckRetry signals to stop.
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -142,11 +142,23 @@ func NewClient() *Client {
...
 // DefaultRetryPolicy provides a default callback for Client.CheckRetry, which
 // will retry on connection errors and server errors.
 func DefaultRetryPolicy(resp *http.Response, err error) (bool, error) {
 	if err != nil {
-		return true, err
+		retryable := true
+		switch t := err.(type) {
+		// net.Error and url.Error have a Temporary() bool that tells us if their
+		// errors are retryable, rather than typecheck for the specific error
+		// classes, we just use that same interface
```
ryanuber: I would probably move this comment to the interface.

client.go
```diff
@@ -184,7 +196,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		checkOK, checkErr := c.CheckRetry(resp, err)
 
 		if err != nil {
-			c.Logger.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
+			c.Logger.Printf("[ERR] %s %s request failed: %v (%T)", req.Method, req.URL, err, err)
```
ryanuber: Is this still useful or was it more for debugging? We don't usually log the error type in our other packages, seems in this case the `Temporary()` should cover us enough to make this not useful.

client_test.go
```diff
@@ -194,6 +194,33 @@ func TestClient_Do_fails(t *testing.T) {
 	}
 }
 
+func TestClient_Do_doesNotRetryOnURISchemeErrors(t *testing.T) {
```
ryanuber: Pretty nit-picky but this could have also been a table test as well. Not a huge deal, I'll let you decide if you want to fold it up or keep the additional test.

client.go
phinze: ✅

client.go
phinze: ✅

client_test.go
phinze: Agreed but decided to 🏈  this for another day.

ryanuber:
@phinze I'd still love to get this merged! There is one test failure (which makes sense!) Let me know if you plan to wrap up or if you want me to take this the last mile. 🏁

phinze:
@ryanuber I've pulled this up 3 or 4 times but haven't found the sustained attention to be able to actually solve the test issue, if you're willing to use your higher degree of loaded go context to fix this up I'd be much obliged! 🙇

ryanuber:
The problem here might be the `Temporary()` method acting strange. It seems that a connection refused error is treated as permanent, though for this use case it really should not be. I dug in a bit, twiddled a few things, and currently have this view of things:

* Trying to use `Temporary()` doesn't work because it returns `false` on connection refused (we wanted `true` so it would retry).
* Doing anything else gets dodgy and we would probably end up doing string comparisons on errors for lots of specific cases (`net/uri` lets lots of things through that `net/http` errors on).

I think we might need to punt on this, given the above.

phinze:
Ok that tracks with my findings. I was hoping you'd see a magical path through but alas it's just a weird story. Going to stand down on this for now. Thanks for the assist!

PR #9, opened 2017-04-21
travis: bump to go 1.8.1

phinze:


ryanuber:
Wondering if we should just `1.8.x` or even the heavy-handed `1.x` this - we probably always want it running against the latest binary Go release.

phinze:
Oh you're here! I was just thinking I was all alone in this so I merged just so I could rebase #8 and leave you a passing build.

phinze:
Yeah I agree - I think `1.x` is fine

PR #10, opened 2017-05-11
I'd like to suppress '[DEBUG]' logs in according to some situation

deadcheat:
i like this package, but i would love to handle print some debug-logs or not.
so i added parameter to Client.

if you don't intend to add feature like this, it's okay,  i use my forked package.

thank you.

### DIFF
--- a/client.go
+++ b/client.go
@@ -42,6 +42,9 @@ var (
 	// We need to consume response bodies to maintain http connections, but
 	// limit the size we consume to respReadLimit.
 	respReadLimit = int64(4096)
+
+	// defaultLogMode is to log both of request and retry
+	defaultDebugLogMode = LogRequestAndRetry
 )
 
 // LenReader is an interface implemented by many in-memory io.Reader's. Used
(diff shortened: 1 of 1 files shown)


ideahitme:
@deadcheat just a passer-by here. I am using this lib too and would like to have this feature. Though I am using a slightly different modification in my form. For example: 

```
        nopLogger := log.New(ioutil.Discard, "", log.LstdFlags)
	c := retryablehttp.NewClient(retryablehttp.WithLogger(nopLogger))
	req, err := retryablehttp.NewRequest("GET", server.URL, nil)
	if err != nil {
		os.Exit(1)
	}

```

This could also help to address another PR for replacing default std log

ryanuber:
It is currently possible to modify the logger used by retryablehttp which gives you access and control over the stream that comes out of it. For example, if you wanted to discard the logs like @ideahitme is doing, you could just do something like:

```go
client.Logger.SetOutput(ioutil.Discard)
```

We also have a log filtering library, [logutils](hashicorp/logutils), which we have used with this library many times. While not quite as flexible, it does a fine job of filtering out messages by their log level, so you could potentially use it with this library to filter out DEBUG-level only if that is what you are after.

ryanuber:
I'm closing this because I think my previous comment should help clarify the intended way for importers to manipulate the log stream produced by this library. Thanks anyways!

PR #11, opened 2017-06-27
Allow customized backoff policies through Client.Backoff() hook

### LINKED ISSUE #12: Allow customizable backoff() function
It'd be nice in some applications to be able to customize the behavior of the backoff logic instead of having it hard coded.

mark-adams:
Previously, a single private backoff() function was used to calculate
the exponential backoff used when a request fails before a retry is
attempted.

It would be useful for developers to be able to customize this backoff
with their own algorithm. This change adds a new Client.Backoff() hook
which defaults to DefaultBackoffPolicy which is equivalent to the
original behavior.

@phinze @ryanuber I was wanting to implement some custom backoff behavior in an application and thought this might be helpful to others in case they wanted to modify the backoff timings. What do you think?

Fixes #12

### DIFF
--- a/client.go
+++ b/client.go
@@ -107,6 +107,11 @@ type ResponseLogHook func(*log.Logger, *http.Response)
 // response body before returning.
 type CheckRetry func(resp *http.Response, err error) (bool, error)
 
+// Backoff specifies a policy for how long to wait between retries.
+// It is called after a failing request to determine the amount of time
+// that should pass before trying again.
+type Backoff func(min, max time.Duration, attemptNum int) time.Duration
+
 // Client is used to make HTTP requests. It adds additional functionality
 // like automatic retries to tolerate minor outages.
 type Client struct {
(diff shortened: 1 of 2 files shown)


mark-adams:
@phinze Would it be possible to get a quick review on this?

Thanks! 😃

client.go
```diff
@@ -159,6 +168,18 @@ func DefaultRetryPolicy(resp *http.Response, err error) (bool, error) {
 	return false, nil
 }
 
+// DefaultBackoffPolicy provides a default callback for Client.Backoff which
+// will perform exponential backoff based on the attempt number and limited
+// by the provided minimum and maximum durations.
+func DefaultBackoffPolicy(min, max time.Duration, attemptNum int) time.Duration {
```
ryanuber: This should probably be named `DefaultBackoff` as the type is simply named `Backoff`.

client.go
```diff
@@ -128,6 +133,9 @@ type Client struct {
 	// CheckRetry specifies the policy for handling retries, and is called
 	// after each request. The default policy is DefaultRetryPolicy.
 	CheckRetry CheckRetry
+
+	// BackoffRetry specifies the policy for how long to wait between retries
```
ryanuber: `Backoff`

mark-adams:
@ryanuber Thanks for the review! 😄 

I think I've made most of the changes that you requested. Would you mind taking another look?

ryanuber:
Sorry for the delay, looks great! Thanks again!

PR #16, opened 2018-01-06
extract status code for logging

bradrf:
It's currently only used for logging later on, but because it's never set, the log never includes the response code (used by https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L237-L239)


ryanuber:
LGTM, thanks!

PR #18, opened 2018-02-10
Allow SetHeader

zhangxichun:
Hi~
In my project, I need set header for JWT


ryanuber:
Hiya @zhangxichun, thanks for the PR!

For this use case I would recommend adding the headers to requests as you make them. To keep things DRY, you could have a wrapper method to do this on each request, or to generate a `*Request` for you with the headers set.

Hope that helps!

zhangxichun:
Oh~ yes.
Thanks reply me.

PR #19, opened 2018-02-13
Added DefaultClient()

cswank:
You can then set the internal defaultClient and call
retryablehttp.Post, for example, with your custom retry
settings.

### DIFF
--- a/client.go
+++ b/client.go
@@ -50,6 +50,12 @@ type LenReader interface {
 	Len() int
 }
 
+// DefaultClient allows you to set the client so your code can
+// call retryablehttp.Post and friends with your settings.
+func DefaultClient(c *Client) {
+	defaultClient = c
+}
+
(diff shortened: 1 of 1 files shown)


ryanuber:
Hey @cswank, thanks for the PR!

When I originally wrote this, I purposely made the default client not settable, as we found a number of security issues in doing so. The problem comes in when multiple libraries using `retryablehttp.Post` and friends are imported into the same project. If the defaults are changed, then those libraries could end up behaving in strange or unexpected ways. This same issue is present in the stdlib `net/http` package, and is the primary motivation behind the `cleanhttp` package produced by us here: https://github.com/hashicorp/go-cleanhttp.

The top-level functions are a convenience for the common case where the defaults are good enough. For any other case I would recommend constructing a client and tracking that separately within your app.

Hope that helps!

cswank:
Cool, thanks for the explanation.

PR #21, opened 2018-04-15
Add Request.WithContext method

### LINKED ISSUE #22: Request with context
Calling WithContext on wrapped [Request](https://godoc.org/github.com/hashicorp/go-retryablehttp#Request), would return *http.Request, so user cannot do:

```go
req, err := NewRequest(http.MethodGet, ts.URL, nil)
if err != nil { /* handle error */ }
ctx, cancel := context.WithCancel(req.Request.Context())
req = req.WithContext(ctx) // compile time error here
```

Also the lib would start retrying on context error, i.e. cancel called. Which is imo wrong. If user calls cancel, the intention is cancel, not retry.

vbauerster:
Calling WithContext on wrapped [Request](https://godoc.org/github.com/hashicorp/go-retryablehttp#Request), would return *http.Request, so user cannot do:

```go
req, err := NewRequest(http.MethodGet, ts.URL, nil)
if err != nil { /* handle error */ }
ctx, cancel := context.WithCancel(req.Request.Context())
req = req.WithContext(ctx) // compile time error here
```

This pull request, fixes #22.

### DIFF
--- a/client.go
+++ b/client.go
@@ -74,6 +74,8 @@ type Request struct {
 	*http.Request
 }
 
+// WithContext returns wrapped Request with a shallow copy of underlying *http.Request
+// with its context changed to ctx. The provided ctx must be non-nil.
 func (r *Request) WithContext(ctx context.Context) *Request {
 	r.Request = r.Request.WithContext(ctx)
 	return r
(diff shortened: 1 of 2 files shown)


alexflint:
@ryanuber I'm also having this issue - would be great to have this in the package

vbauerster:
Rebased to master, with extra one commit.

jefferai:
Thanks!

PR #23, opened 2018-04-30
Drop request wrapper

### LINKED ISSUE #22: Request with context

alexflint:
This fixes #22 in a different way to the current proposed fix (#21):
- Do() can now take any http.Request as input. If the request body is not seekable then the contents of the request are read to a buffer, and `bytes.Reader` is used, which is seekable.
  - but if the request body is _already_ seekable then the copy is not made
  - and in my experience almost all request bodies are already in-memory buffers created by marshaling json / xml / text / etc, so I don't think this is a big loss
  - and realistically, if the user passes in a request body that was not seekable, then under the old API the user just would have done the in-memory copy trick anyway
  - so it really does not seem like we're losing a lot here
- drops the Request wrapper struct (converts it to a type alias for http.Request for backwards compatbility)
- the `NewRequest` function is now just a direct passthrough to `http.newRequest`

One of the nice things about this approach is that it allows users of this library to define interfaces that can match both `http.Client` and `retryablehttp.Client`, so e.g. that retryable HTTP requests can be dependency-injected into applications:
```
type MyHttpClient interface {
  Do(req *http.Request) (*http.Response, error)
}

func doSomething(client MyHttpClient) { ... }

func main() {
  if (shouldBeRetryable) {
    doSomething(&retryablehttp.Client{...})
  } else {
   doSomething(&http.Client{...})
  }
}
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -173,6 +173,11 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	c.Logger.Printf("[DEBUG] %s %s", req.Method, req.URL)
 
 	if req.Body != nil {
+		// below we will prevent net/http from closing the request body after
+		// the first try, which means we must take responsibility for closing it
+		// after we are done retrying
+		defer req.Body.Close()
+
 		// if body is not rewindable then read it to a buffer that is rewindable
 		if readSeeker, ok := req.Body.(io.ReadSeeker); ok {
 			// always override any underlying Close function on req.Body so that
@@ -191,7 +196,7 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 		var code int // HTTP response code
 
 		// Always rewind the request body when non-nil.
-		if req.Body != nil {
+		if req.Body != nil && i > 0 {
 			// this type cast will always succeed due to the steps above
 			if _, err := req.Body.(io.Seeker).Seek(0, 0); err != nil {
 				return nil, fmt.Errorf("failed to seek body: %v", err)
--- a/client.go
+++ b/client.go
@@ -57,7 +57,7 @@ type Request = http.Request
 
 // NewRequest creates a new HTTP request. It returns http.NewRequest and
 // exists only for backwards compatibility.
-func NewRequest(method, url string, body io.ReadSeeker) (*Request, error) {
+func NewRequest(method, url string, body io.Reader) (*Request, error) {
 	return http.NewRequest(method, url, body)
 }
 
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -180,26 +157,53 @@ func DefaultBackoff(min, max time.Duration, attemptNum int, resp *http.Response)
...
 // Do wraps calling an HTTP method with retries.
-func (c *Client) Do(req *Request) (*http.Response, error) {
+func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	c.Logger.Printf("[DEBUG] %s %s", req.Method, req.URL)
 
+	if req.Body != nil {
+		// if body is not rewindable then read it to a buffer that is rewindable
+		if readSeeker, ok := req.Body.(io.ReadSeeker); ok {
+			// always override any underlying Close function on req.Body so that
+			// net/http does not close it after the first try
```
jefferai: If this is preventing net/http from closing after the first attempt, doesn't it also mean that the body will now never be closed? I would think you'd want to defer a close on the underlying body before wrapping.

ryanuber:
Hey @alexflint I do like that this sheds a layer of abstraction - nice! In general I think this will be fine. I would like if we could add some notes to the README / godoc to be clear about what happens if the body is not seekable. It is probably due for an update as a result of this change anyways :)

bernielomax:
Any chance this PR can be finalized and merged?

alexflint:
@ryanuber ping

ryanuber:
FYI over in #29, @jefferai is working on an issue where it appears that any underlying `io.Seeker` being `Seek()`'ed and reused might not be a reliable option anymore.  We are still thinking about this at this point but the two PR's are related so wanted to provide the context here.

crisguitar:
Hello, we are in the need of this, any chance you can merge the PR?

jefferai:
This was superceded by #29, just forgot to close it. Sorry about that!

PR #24, opened 2018-05-09
Check that logger is not nil before logging

jefferai:


jefferai:
Should be uncontroversial; merging

PR #25, opened 2018-05-09
Add ErrorHandler, LinearJitterBackoff, and some minor fixes

jefferai:
* Adds an ErrorHandler that is invoked with the response and error of
the last retry; default is prior behavior

* Added LinearJitterBackoff, which uses min/max as a min/max for jitter
calculations and multiplies by attempt num

* Don't drain the response body if we're not going to retry, just close

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"log"
 	"math"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"os"
@@ -112,6 +113,12 @@ type CheckRetry func(resp *http.Response, err error) (bool, error)
 // that should pass before trying again.
 type Backoff func(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration
 
+// ErrorHandler is called if retries are expired, containing the last status
+// from the http library. The default will close the body and return an error
+// indicating how many tries were attempted. If overriding this, be sure to
+// close the body if needed.
+type ErrorHandler func(resp *http.Response, err error, numTries int) (*http.Response, error)
+
 // Client is used to make HTTP requests. It adds additional functionality
 // like automatic retries to tolerate minor outages.
 type Client struct {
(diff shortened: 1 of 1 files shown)


jefferai:
It's semi urgent. I have about 100 tests in the Vault Go SDK though that don't work without it :-)

client.go
```diff
@@ -112,6 +113,12 @@ type CheckRetry func(resp *http.Response, err error) (bool, error)
 // that should pass before trying again.
 type Backoff func(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration
 
+// ErrorHandler is called if retries are expired, containing the last status
+// from the http library. The default will close the body and return an error
+// indicating how many tries were attempted. If overriding this, be sure to
+// close the body if needed.
+type ErrorHandler func(resp *http.Response, err error, numTries int) (*http.Response, error)
```
jbardin: My opinion is that we don't need this, and can return the response and error directly. That pattern isn't unprecedented, since net/http already does that for failed redirects. It also shouldn't break any existing code, since there is no guarantees made about the first value in the presence of an error. 

I'm not opposed to this addition if there a solid reason, but I need some convincing ;)

client.go
```diff
@@ -112,6 +113,12 @@ type CheckRetry func(resp *http.Response, err error) (bool, error)
 // that should pass before trying again.
 type Backoff func(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration
 
+// ErrorHandler is called if retries are expired, containing the last status
+// from the http library. The default will close the body and return an error
```
jbardin: I don't see a default set anywhere.

jefferai: I've updated the comment to mean default behavior rather than handler

client.go
jefferai: Really, it's just for backwards compat. I'm not opposed to making that the default behavior if we just want to force others to get in line as I think it makes sense.

PR #29, opened 2018-05-30
Allow various types of input to avoid racy seeking

jefferai:
With the previous code, some data races have very occasionally been
observed in client code due to the seek call (the write) and a read call
that is likely coming from deep in the bowels of the net/http package as
for HTTP/2 connections the request body reads happen in goroutines since
it's multiplexed. The exact cause is not yet known, but the same
behavior has not been observed in Pester, which copies the request body
anew each time instead of seeking. Even if there is not technically any
concurrent modification, a read after an out-of-order seek can mean
unexpected data can be read.

This eliminates that possibility in many cases by taking advantage of
the fact that we always seek to the first byte and simply eliminating
the need for seeks entirely when possible, instead using new readers
over the same underlying byte slice.

We accept various different ways of inputting the data and try to do the
right thing in all cases. io.ReadSeeker is still supported (but
potentially racy).

### DIFF
--- a/client.go
+++ b/client.go
@@ -58,7 +58,7 @@ type LenReader interface {
 type Request struct {
 	// body is a seekable reader over the request body payload. This is
 	// used to rewind the request data in between retries.
-	body func() io.Reader
+	body func() (io.Reader, error)
 
 	// Embed an HTTP request directly. This makes a *Request act exactly
 	// like an *http.Request so that all meta methods are supported.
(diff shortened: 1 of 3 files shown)


jefferai:
There's one thing to think about in particular:

If the ReadSeeker that was coming in was coming from, say, an os.File, IIRC Go can be super efficient about reading directly from the file onto the wire. So the code that's there now that firts reads it all into a buffer will destroy that efficiency in the name of potential safety. So the question is, do we "buyer beware" it if they pass in a ReadSeeker, while adding more safety when the initial data is coming from an in-memory byte slice already.

client.go
```diff
@@ -249,9 +311,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 
 		// Always rewind the request body when non-nil.
 		if req.body != nil {
-			if _, err := req.body.Seek(0, 0); err != nil {
-				return nil, fmt.Errorf("failed to seek body: %v", err)
-			}
+			req.Request.Body = ioutil.NopCloser(req.body())
```
ryanuber: I think we need to put the `NopCloser` wrapper specifically in the `ReadSeeker` case only. We want it to close any other kind of reader.

client.go
```diff
@@ -55,32 +58,91 @@ type LenReader interface {
 type Request struct {
 	// body is a seekable reader over the request body payload. This is
 	// used to rewind the request data in between retries.
-	body io.ReadSeeker
+	body func() io.Reader
```
ryanuber: This would probably be better as `func() (io.Reader, error)`, so if you are doing something like opening a file you have an out in error cases. You could pass `func() (io.Reader, error) { return os.Open("foo.txt") }` then, for example.

client.go
jefferai: I had it like that but wasn't really sure. The reason is that if someone passes in a func that returns an io.Reader, that io.Reader may also implement a Close method so the check in net/http for whether it satisfies io.Closer will pass. IOW without knowing the full spectrum of things that might be passed in, this seemed safer. But we could leave that up to the user to sort out.

client.go
jefferai: Fair enough, I'll change that.

client.go
ryanuber: Yeah, I think with the supported types we have here, the only time we will want to prevent the `Close()` will be for the `ReadSeeker`, since the types from the `bytes` package don't implement a `Close()`, and the function type definitely needs to have its readers closed if they support it.

Just for example's sake, if I pass a function that returns an `os.File`, that will never be closed and leak an FD if we mask the `Close()` method.

jefferai: Yep, sounds good, will fix.

PR #32, opened 2018-11-10
Added interface for custom loggers

### LINKED ISSUE #6: Allow pluggable Logrus logger
The retryablehttp lib is currently forcing us to use the default go.Logger.
It would be easy to allow other loggers (like Logrus) by replacing your use of `*Logger` by an interface. For instance look at this:
https://github.com/sirupsen/logrus/blob/master/logrus.go#L97

harnash:
This interface should allow to plug-in custom loggers
more easily.
 
Fixes #6

### DIFF
--- a/client.go
+++ b/client.go
@@ -183,18 +183,24 @@ func NewRequest(method, url string, rawBody interface{}) (*Request, error) {
 	return &Request{body, httpReq}, nil
 }
 
+// Logger interface allows to use other loggers than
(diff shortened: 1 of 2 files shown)


jefferai:
@ryanuber Can we revert this? This is currently not an interface satisfiable by go-hclog. If we're going to make it an interface it would be nice to at least figure out first-party support.

PR #33, opened 2018-11-12
Use a math.Rand variable for pseudo-random numbers

svanharmelen:
While initializing a new `*math.Rand` variable each time the backoff function is called is not the most expensive task, it isn’t needed if we create a single private variable containing a properly seeded `*math.Rand` instance.

### DIFF
--- a/client.go
+++ b/client.go
@@ -49,6 +49,9 @@ var (
 	// a new client. It is purposely private to avoid modifications.
 	defaultClient = NewClient()
 
+	// random is used to generate pseudo-random numbers.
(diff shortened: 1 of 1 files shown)


jefferai:
Yeah, it's fine by me.

jefferai:
I should say that I also don't really see any need for this unless it can be shown that there is some measurable impact that seeding it each time is having. So, either way.

svanharmelen:
I don't know what the actual measurable impact, but it seems unneeded to generates a couple of structs and do some calculations (to set the seed) for every backoff calculation for every HTTP call that is retried. So in that case I'll go ahead and merge this one.

Thanks!

PR #35, opened 2018-12-04
Fix potential data race, introduced by rand.NewSource

vbauerster:
rand.NewSource doc clearly says:
Unlike the default Source used by top-level functions, this source is not safe for concurrent use by multiple goroutines.

### DIFF
--- a/client.go
+++ b/client.go
@@ -29,14 +29,14 @@ import (
 	"io/ioutil"
 	"log"
 	"math"
-	"math/rand"
 	"net/http"
 	"net/url"
 	"os"
 	"strings"
 	"time"
 
-	"github.com/hashicorp/go-cleanhttp"
+	cleanhttp "github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/go-retryablehttp/saferand"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


jefferai:
Should we just revert #33 instead?

mmatczuk:
Hi @jefferai @vbauerster,
I'm curious why do you think that taking a mutex is a better idea than calling rand.New? Do you have benchmark data that you can share?

jefferai:
@mmatczuk I don't. I was neutral on #33 for exactly that reason, there wasn't any demonstrated need other than theoretical optimization. That's why I suggested reverting #33 to fix this.

I would guess any time lost creating a new rand source is totally dwarfed by the actual backoff time. Personally I think we should just revert #33.

@jbardin

vbauerster:
@mmatczuk @jefferai 
[](https://postimg.cc/JtSDfRn0)

jefferai:
Okay so we're talking 10us against a sleep in the range of hundreds of milliseconds to tens of seconds. I say we revert #33.

vbauerster:
Here same code with concurrent benchmarks:
[](https://postimg.cc/cKQBHT9q)

jbardin:
I don't think the time alone here is a concern, but memory can be since each `rand.Source` contains a copy of a precooked array which is around 4k.

On the other hand, this PRs code already exists in the math/rand package as the default source, which is also safe for concurrent calls to `Seed` and `Read`. Since this is only for randomized jitter, I would elect to revert 33, and simply ensure the global source is seeded.

jefferai:
Closing in deference of someone putting together saferand

PR #38, opened 2018-12-12
Fix race condition in rand.Source

jbardin:
This reverts #33, fixing the data race, and removes the excess allocations by using the global random source which is safe for concurrent use. 

An alternative PR to #35

### DIFF
--- a/client.go
+++ b/client.go
@@ -49,9 +49,6 @@ var (
 	// a new client. It is purposely private to avoid modifications.
 	defaultClient = NewClient()
 
-	// random is used to generate pseudo-random numbers.
-	random = rand.New(rand.NewSource(time.Now().UnixNano()))
-
 	// We need to consume response bodies to maintain http connections, but
 	// limit the size we consume to respReadLimit.
 	respReadLimit = int64(4096)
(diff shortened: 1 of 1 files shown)


jbardin:
I figure if you're using the global source, you can't count on it not being seeded already, but I see your point. If we're worried out the global source, then I think 35 is fine, or even putting that into a discoverable package, .e.g "go-cleanrandom" (which might also benefit from a `crypto/rand` backed `rand.Source`, which comes in handy sometimes too) ;)

jefferai:
> I figure if you're using the global source, you can't count on it not being seeded already

If you are in your application, I think you should be able to assume that you can control the global source, rather than libraries like this one are assuming they should have unfettered control instead of your application  :-)

I'm into the cleanrandom idea that has a couple of different concurrent-safe and/or crypto/rand backed possibilities. In the meantime, I don't think there's been any credible/demonstrable speed issues, and theoretical speed issues were the only reason for #33 in the first place. I vote we revert #33 to get rid of the data race, then anyone interested can take on `go-cleanrandom`.

jbardin:
SGTM -- that way we don't have `saferand` duplicated under the unrelated repo 👍

jefferai:
Closing in deference of someone putting together saferand

PR #47, opened 2019-03-19
Feature/metrics

influx6:
PR adds prometheus metric gathering for a http client.


influx6:
I added a `RequestModifier` to the types available to a `Client`, to allow addition of new headers or modification of request context as desirable.

PR #48, opened 2019-03-22
Switch expr simplification

moeryomenko:
Signed-off-by: Maxim Eremenko <moeryomenko@gmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -106,14 +106,14 @@ func (r *Request) BodyBytes() ([]byte, error) {
 // NewRequest creates a new wrapped request.
 func NewRequest(method, url string, rawBody interface{}) (*Request, error) {
 	var err error
-	var body ReaderFunc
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -110,10 +110,10 @@ func NewRequest(method, url string, rawBody interface{}) (*Request, error) {
 	var contentLength int64
 
 	if rawBody != nil {
-		switch rawBody.(type) {
+		switch rawBody := rawBody.(type) {
```
jefferai: To make it clearer what's happening here (so people's minds don't miss the scope change even if the compiler doesn't), can you make the variable you're assigning to named something else?

moeryomenko:
ping @jefferai

jefferai:
@moeryomenko Unfortunately you deleting the repo means I can't merge this. Can you restore it?

PR #49, opened 2019-03-22
Rework logger for support level logging

moeryomenko:


jefferai:
I don't think this will be merged. Not only is it backwards incompatible but forcing logging prefixes is something we've explicitly gotten away from. If you have specific needs I suggest filing an issue with your use-cases and we can see if they can be accommodated.

PR #50, opened 2019-03-23
Cancel the retry sleep if the request context is cancelled.

snabb:


client.go
```diff
@@ -445,7 +445,12 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		if c.Logger != nil {
 			c.Logger.Printf("[DEBUG] %s: retrying in %s (%d left)", desc, wait, remain)
 		}
-		time.Sleep(wait)
+		// The following is equivalent to time.Sleep(wait) with support for context cancellation.
```
jefferai: Please remove this comment

snabb: PTAL

client.go
```diff
@@ -445,7 +445,11 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		if c.Logger != nil {
 			c.Logger.Printf("[DEBUG] %s: retrying in %s (%d left)", desc, wait, remain)
 		}
-		time.Sleep(wait)
+		select {
+		case <-req.Request.Context().Done():
```
jefferai: This can just be `req.Context().Done()` and the line under can be `req.Context().Err()`.

snabb: Ok I will change it. I did it this way because I saw that this is the style used elsewhere in the code. I assumed it is the project's preferred coding style to be explicit about the embedded struct. For example:
- https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L405
- https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L388
- etc.

snabb: Done. PTAL.

jefferai: I think those examples are outliers. I didn't see those but I saw plenty of places e.g. `req.URL` was used directly. I'd say be consistent. if you're going to change this, maybe you can change the other places too.

snabb: Ok, I will change the others to match.

snabb: Done. PTAL.

PR #56, opened 2019-05-07
Return response in error case

### LINKED ISSUE #46: Return response in error case
Right now the retryable client will return nil and the error if a request failed. It would be good to still have access to the response to for example check the status code or any error messages. 

I know I can define a custom error handler to do this, but I think it should be the default behavior.

valentin-krasontovitsch:
We follow the default behavior of the net/http package and return the
response in case of an error.

Closes #46

### DIFF
--- a/client.go
+++ b/client.go
@@ -456,12 +456,8 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		return c.ErrorHandler(resp, err, c.RetryMax+1)
 	}
 
-	// By default, we close the response body and return an error without
-	// returning the response
-	if resp != nil {
-		resp.Body.Close()
-	}
-	return nil, fmt.Errorf("%s %s giving up after %d attempts",
+	// By default, we return the response and an error
+	return resp, fmt.Errorf("%s %s giving up after %d attempts",
 		req.Method, req.URL, c.RetryMax+1)
 }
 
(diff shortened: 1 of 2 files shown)


jefferai:
This is kind of a big breaking change, which is why ErrorHandler exists. If we do this, suddenly any user not already checking for nil and closing the response body if not (and there's no reason to do this since we have defined the response to be nil) will skip out on closing response bodies which will eventually cause resource issues.

I don't think we can make this change, at least not without a major version bump (to v2).

valentin-krasontovitsch:
That is unfortunately a very good point : )

So we'll have to wait for a new major release, I suppose...

Is there in general agreement that this is a desired change though?

jefferai:
IMHO, not enough to warrant actually cutting a new major version, since you can access it from the ErrorHandler. Maybe the method just needs better commenting about the use of ErrorHandler?

elithe1:
I had the same problem because when using the regular net/http client i used the responses migrating to retryablehttp felt weird not having them.
After reading this discussion and using PassthroughErrorHandler I get my expected behaviour :)

jefferai:
Ah good point, I actually forgot about PassthroughErrorHandler, which was written for this exact use case. OP, please use that :-D

valentin-krasontovitsch:
sorry for the late reply: wasn't implying that we should do a new major release just because of this, by all means : ) was just wondering what the general opinion was about this change - whether it's something that you could see become part of the package, or not.

Definitely gonna check out the workaround / suggested method! Thanks for the pointer ^^

PR #57, opened 2019-05-07
Use cleanhttp.DefaultPooledConnection and enforce cleanup

### LINKED ISSUE #44: Mention use of cleanhttp which by default disables keepalives
It does not concern in most use cases but users should be aware that keep-alives are disabled explicitly by default. Most people would assume defaults and expect them enabled.

ie: I was planning to use it for an HTTP consumer and it definitely can re-use underlying connection. I only got to know while I was going thru source code.

What I mean is library should highlight the fact what it is doing internally, so that users are aware and can make an informed decision. Happy to make a PR :)

jefferai:
This allows reuse of the existing connection on retries, since they're
going to the same host, but we explicitly close idle connections to
ensure they don't stick around forever after the series of requests is
done.

Fixes #44

### DIFF
--- a/client.go
+++ b/client.go
@@ -277,7 +277,7 @@ type Client struct {
 // NewClient creates a new Client with default settings.
 func NewClient() *Client {
 	return &Client{
-		HTTPClient:   cleanhttp.DefaultClient(),
+		HTTPClient:   cleanhttp.DefaultPooledClient(),
 		Logger:       log.New(os.Stderr, "", log.LstdFlags),
 		RetryWaitMin: defaultRetryWaitMin,
 		RetryWaitMax: defaultRetryWaitMax,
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -422,6 +427,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 			if checkErr != nil {
 				err = checkErr
 			}
+			c.HTTPClient.CloseIdleConnections()
```
evanphx: This shouldn't be here. This branch is when good requests are returned and closing idle here will make it mostly useless. You could probably safely move this up 2 lines inside check checkErr if so it's only run if we're running with an error.

jefferai: Why do we not want to close idle connections if it's successful?

Before this patch we have keepalives totally disabled. Closing idle connections in all cases preserves that behavior except when actively retrying.

evanphx: I thought the point of this patch was to leave connections alive so that future ones are faster?

jefferai: Yes, but scoped only to the library's own retries.

evanphx: Ah ok.

bodhi: This breaks compatibility with Go < v1.12, `HTTPClient.CloseIdleConnections` was [only added in v1.12](https://golang.org/doc/go1.12#net/http) 😿

PR #60, opened 2019-06-26
adds debug and error methods to logger interface

matlockx:
This change makes it easier to use a different log implementation and use the log level from these loggers. I left out other additional ones (Infof, etc) since they are not used here

### DIFF
--- a/client.go
+++ b/client.go
@@ -225,7 +225,29 @@ func NewRequest(method, url string, rawBody interface{}) (*Request, error) {
 // Logger interface allows to use other loggers than
 // standard log.Logger.
 type Logger interface {
+	Debugf(string, ...interface{})
 	Printf(string, ...interface{})
+	Errorf(string, ...interface{})
+}
+
+type logger struct {
+	stdLogger *log.Logger
+}
+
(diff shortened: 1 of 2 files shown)


danlamanna:
Is there anything the community can do to expedite getting this merged? It makes integrating with other packages difficult.

jefferai:
I haven't had any internal conversations around this but my gut feeling is that we're more likely to change (since it's < v1) or v1 the package and use an hclog Log, which already has methods for pulling out std loggers. @evanphx @ryanuber thoughts?

matlockx:
I'd be fine with that as well as long as it's going to happen somewhen near ;)

evanphx:
Adding hclog into here is probably the direction to take. In terms of not breaking the existing API, detecting that `Logger` is nil and using either a `HCLogger` value or `hclog.L()` is the right direction. Or we could just break API entirely and change Logger to `hclog.Logger`.

jefferai:
Given it's a 0.x version I'm a fan of the latter option if it results in a cleaner API/experience.

evanphx:
I'll write up the PR today.

PR #62, opened 2019-07-30
Allow Logger value to use hclog or previous Logger interface

evanphx:
Related to #60 this changes the Logger to be hclog.Logger with the smallest API change possible.

### DIFF
--- a/client.go
+++ b/client.go
@@ -27,16 +27,15 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"log"
 	"math"
 	"math/rand"
 	"net/http"
 	"net/url"
-	"os"
 	"strings"
 	"time"
 
 	cleanhttp "github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/go-hclog"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


jefferai:
I'm not sure about this approach -- it's an API break but it's also breaking the format of the text which may cause some havoc for those expecting a particular format.

I wonder if instead we should just make the logger that comes in an `interface{}` and add support for it being an `hclog.Logger` and using that natively but falling back to the existing interface (maybe renamed `DeprecatedLogger` or something). That will mean that we won't break client code but anything using hclog already can easily just pass that into the client.

I don't really want to end up with a ton of PRs asking for support for this and that interface so I think we could take a stance that it's just hclog and the deprecated one. And maybe a third one being a stdlog.

WDYT?

evanphx:
@jefferai Ah, I guess I misunderstood what you were saying in #60. I thought you were saying that you were totally fine breaking the api because it was not yet 1.0 and folks can just pin to earlier versions via go.mod.

I'm fine using some interface casting tricks to allow multiple concrete types, that's not a biggy. But changing the text format output is mostly a non-starter for hclog, so we'd need to be ok with that change.

jefferai:
That was what I was suggesting but I didn't realize that this package was using hardcoded level strings. So it's not just an API change but also an output change.

matlockx:
any news here? i really would like to get rid of the debug post requests in my logs

danlamanna:
Should the case of the default logger use Errorf and Debugf instead of printf?

matlockx:
can we get a new release please?

jefferai:
Would like to see first if @evanphx has thoughts on debugf and errorf as asked above.

evanphx:
Changing the definition of `Logger` was a non-goal of this work so that can be considered as a separate PR.

evanphx:
Just made a release.

PR #67, opened 2019-09-11
feat: allow access to LogHooks with nil Logger

therealjt-abstract:
# Summary

* Allow `RequestLogHook` and `ResponseLogHook` to be accessed even if the `client.Logger` is `nil`
* Add tests for ^

# Use Case

We're using a logging package that doesn't conform to `Logger` interface, but still want access to values available through `LogHook`s. Specifically,  we're using `RequestLogHook` to access the retry count and pass it to our own internal structured JSON logging package.

However, setting `Client.Logger = nil` (in order to suppress logging to STDOUT) skips the block of code where `RequestLogHook` accesses the retry count ([here](https://github.com/hashicorp/go-retryablehttp/blob/v0.6.2/client.go#L457)). The presence of the `default` case in the `switch` [here](https://github.com/hashicorp/go-retryablehttp/blob/v0.6.2/client.go#L463) makes me think that this behavior (suppressed logging, but still able to access values in `LogHook`s) was meant to be supported.

# Considered Alternatives

The alternative of disabling the logger with a noop logger or other similar approach has a small performance cost (see [here](https://stackoverflow.com/questions/10571182/how-to-disable-a-log-logger/34457930#34457930), tho the benchmarks are probably out of date, the idea that it will cost _something_ to have a noop logger is what is most important)

### DIFF
--- a/client.go
+++ b/client.go
@@ -457,13 +457,13 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 			}
 		}
 
-		if c.RequestLogHook != nil && logger != nil {
+		if c.RequestLogHook != nil {
 			switch v := logger.(type) {
 			case Logger:
 				c.RequestLogHook(v, req.Request, i)
 			case hclog.Logger:
 				c.RequestLogHook(hookLogger{v}, req.Request, i)
-			default:
+			case nil:
 				c.RequestLogHook(nil, req.Request, i)
 			}
 		}
--- a/client.go
+++ b/client.go
@@ -463,7 +463,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 				c.RequestLogHook(v, req.Request, i)
 			case hclog.Logger:
 				c.RequestLogHook(hookLogger{v}, req.Request, i)
-			case nil:
+			default:
 				c.RequestLogHook(nil, req.Request, i)
 			}
 		}
@@ -494,7 +494,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 					c.ResponseLogHook(v, resp)
 				case hclog.Logger:
 					c.ResponseLogHook(hookLogger{v}, resp)
-				case nil:
+				default:
 					c.ResponseLogHook(nil, resp)
 				}
 			}
--- a/client.go
+++ b/client.go
@@ -47,6 +47,9 @@ var (
 	defaultRetryWaitMax = 30 * time.Second
 	defaultRetryMax     = 4
 
+	// defaultLogger is the logger provided with defaultClient
+	defaultLogger = log.New(os.Stderr, "", log.LstdFlags)
+
 	// defaultClient is used for performing requests without explicitly making
 	// a new client. It is purposely private to avoid modifications.
 	defaultClient = NewClient()
@@ -309,7 +312,7 @@ type Client struct {
 func NewClient() *Client {
 	return &Client{
 		HTTPClient:   cleanhttp.DefaultPooledClient(),
-		Logger:       log.New(os.Stderr, "", log.LstdFlags),
+		Logger:       defaultLogger,
 		RetryWaitMin: defaultRetryWaitMin,
 		RetryWaitMax: defaultRetryWaitMax,
 		RetryMax:     defaultRetryMax,
--- a/client.go
+++ b/client.go
@@ -477,27 +477,25 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		// Check if we should continue with retries.
 		checkOK, checkErr := c.CheckRetry(req.Context(), resp, err)
 
-		if logger != nil {
-			if err != nil {
+		if err != nil {
+			switch v := logger.(type) {
+			case Logger:
+				v.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
+			case hclog.Logger:
+				v.Error("request failed", "error", err, "method", req.Method, "url", req.URL)
+			}
+		} else {
+			// Call this here to maintain the behavior of logging all requests,
+			// even if CheckRetry signals to stop.
+			if c.ResponseLogHook != nil {
+				// Call the response logger function if provided.
 				switch v := logger.(type) {
 				case Logger:
-					v.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
+					c.ResponseLogHook(v, resp)
 				case hclog.Logger:
-					v.Error("request failed", "error", err, "method", req.Method, "url", req.URL)
-				}
-			} else {
-				// Call this here to maintain the behavior of logging all requests,
-				// even if CheckRetry signals to stop.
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -454,13 +457,13 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
...
 
-		if c.RequestLogHook != nil && logger != nil {
+		if c.RequestLogHook != nil {
 			switch v := logger.(type) {
 			case Logger:
 				c.RequestLogHook(v, req.Request, i)
 			case hclog.Logger:
 				c.RequestLogHook(hookLogger{v}, req.Request, i)
-			default:
+			case nil:
```
therealjt-abstract: I have a personal preference for being super explicit, but have no problem with changing this to `default` if you prefer that. Same for [below](https://github.com/hashicorp/go-retryablehttp/pull/67/files#diff-d0b7f86cea13f90d3cc0cc2338da8d5fR497)

therealjt-abstract:
Just wanted to check in to see if you had any feedback on this

jefferai:
It seems like you're doing two different things here: one, removing the check for the logger being nil; two, changing how formatting works for some of the logging statements.

Please split this into two separate PRs. I'm not sure the first will be merged as it's not really hard to create a noop logger and I don't think a single benchmark on a single platform of a two year old Go version is really relevant. Generally speaking the amount of time it will take to execute an HTTP request, have it fail, possibly backoff, possibly marshal parameters and write to an output device to log, and finally retry the whole thing again is going to absolutely dwarf the amount of time it takes to instantiate a struct, even if it's tens or hundreds of nanoseconds.

client.go
```diff
@@ -474,27 +477,25 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		// Check if we should continue with retries.
 		checkOK, checkErr := c.CheckRetry(req.Context(), resp, err)
 
-		if logger != nil {
-			if err != nil {
+		if err != nil {
+			switch v := logger.(type) {
+			case Logger:
+				v.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
```
therealjt-abstract: This line wasn't removed like the unified diff makes it appear, if you view this PR in Split mode it becomes clearer I just moved the blocks around when I removed the `if logger != nil! check.

client.go
```diff
@@ -474,27 +477,25 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
...
+				v.Error("request failed", "error", err, "method", req.Method, "url", req.URL)
+			}
+		} else {
+			// Call this here to maintain the behavior of logging all requests,
+			// even if CheckRetry signals to stop.
+			if c.ResponseLogHook != nil {
+				// Call the response logger function if provided.
 				switch v := logger.(type) {
 				case Logger:
-					v.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
```
therealjt-abstract: This line has not been actually modified, it still exists in the same form, see https://github.com/hashicorp/go-retryablehttp/pull/67/files#diff-d0b7f86cea13f90d3cc0cc2338da8d5fR483

client.go
```diff
@@ -474,27 +477,25 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
...
+					c.ResponseLogHook(v, resp)
 				case hclog.Logger:
-					v.Error("request failed", "error", err, "method", req.Method, "url", req.URL)
```
therealjt-abstract: This line has not actually been modified, it still exists in the same form, see https://github.com/hashicorp/go-retryablehttp/pull/67/files#diff-d0b7f86cea13f90d3cc0cc2338da8d5fR485

client.go
```diff
@@ -47,6 +47,9 @@ var (
 	defaultRetryWaitMax = 30 * time.Second
 	defaultRetryMax     = 4
 
+	// defaultLogger is the logger provided with defaultClient
+	defaultLogger = log.New(os.Stderr, "", log.LstdFlags)
```
therealjt-abstract: I only added this to keep the code DRY and re-usable since I used the same logger in the tests (see [here](https://github.com/hashicorp/go-retryablehttp/pull/67/files#diff-51565ad50d7426213c2f33579f6f0e71R299)). Happy to copy and paste if you prefer that

jefferai:
Cost savings have to be traded off against whatever complexity and/or risk they provide. In this case saving a few nanoseconds, or even a few hundred nanoseconds, off an operation that is likely to take between tens of microseconds at the absolute bare minimum and seconds to minutes at max isn't super persuasive as a reason to make a change on its own.

That said, I think there are other reasons to consider this. I don't think this change is sufficient as it is but for a different reason: the underlying logger passed in is an `interface{}` but if it's not a type we know about we panic in one location and change it to nil in others. I wonder if in addition to allowing nil -- or quite possibly instead of -- we should remove the panic entirely, and pass whatever we _do_ have to the hook function. That way you can pass arbitrary loggers that don't satisfy our interfaces but where in the hook function you can type assert to whatever your underlying type is.

@evanphx thoughts?

client.go
evanphx: I'd prefer it be default in case someone passes a weird logger it does the same thing.

client.go
```diff
@@ -474,27 +477,25 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
...
-					switch v := logger.(type) {
-					case Logger:
-						c.ResponseLogHook(v, resp)
-					case hclog.Logger:
-						c.ResponseLogHook(hookLogger{v}, resp)
-					default:
-						c.ResponseLogHook(nil, resp)
-					}
+					c.ResponseLogHook(hookLogger{v}, resp)
+				case nil:
```
evanphx: Should use default instead of nil here in the offhand case that someone passes a strange logger.

evanphx:
I do think that passing the logger value as-is to the hooks is probably the most ergonomic but it breaks the ABI so we probably shouldn't. In that case, the user can just ignore the argument and use whatever logger they have constructed elsewhere (perhaps installing a closure into the hook functions).

client.go
therealjt-abstract: Cool, changed!

client.go
therealjt-abstract: Changed

therealjt-abstract:
I went down the route you metioned @evanphx and installed a closure in the hook function to pass values to our logger and it works really nicely, all that was left was gaining the ability to turn off the extra logging to stdout to make it perfect.

jefferai:
@evanphx Ah I forgot that LogHook doesn't take in an interface, it takes in a Logger.

therealjt-abstract:
@evanphx Got those changes you requested in last week, let me know if you have any other feedback

evanphx:
This looks fine to me. @jefferai any objection to merging?

jefferai:
@evanphx Nope all good

therealjt-abstract:
Great, thanks for the feedback all

PR #70, opened 2019-10-11
Propogate Do() HTTP error up when retries are exceeded

### LINKED ISSUE #69: HTTP Request error should be propagated to the caller
Hello,

I'm working to add retryablehttp to a CLi I'm building for something, and noticed an interesting change in error messages as a result of doing that. Specifically when the `Do()` method exceeds its number of retries, the error it returns does not include the original error that was encountered:

https://github.com/hashicorp/go-retryablehttp/blob/46f7ba56794607f00b2e4797b3090b76dff6791e/client.go#L556-L557

So if there was a connection reset error causing it, versus an HTTP 500, the error doesn't provide that context to the consumer.

You do provide an `ErrorHandler`, but with what you've provided to that handler it's impossible to create an error string that is identical to the one above. This is because the `Do()` method relies on the `*http.Request` being in-scope to generate the retries exceeded error string, but that's not passed to `ErrorHandler`.

I'm interested in seeing whether: 

1. We can enhance that line of code above to also include the original error string in the message.
2. The `ErrorHandler` can be updated to receive an `*http.Request` so that we can create an error identical to that of `Do()`.

theckman:
Add propagation of the HTTP request/response error up the call stack when the
number of retries are exceeded. Because the ErrorHandler does not receive a copy
of the request, it's impossible to use the ErrorHandler to generate an enhanced
version of the default Retries Exceeded error message.

This change takes the non-API breaking approach to solving this, by including
the error string in the returned error (if an error has occured).

Fixes #69

### DIFF
--- a/client.go
+++ b/client.go
@@ -432,6 +432,48 @@ func DefaultRetryPolicy(ctx context.Context, resp *http.Response, err error) (bo
 	return false, nil
 }
 
+// ErrorPropagatedRetryPolicy is the same as DefaultRetryPolicy, except it
+// propagates errors back instead of returning nil. This allows you to inspect
+// why it decided to retry or not.
+func ErrorPropagatedRetryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	// do not retry on context.Canceled or context.DeadlineExceeded
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+
+	if err != nil {
+		if v, ok := err.(*url.Error); ok {
+			// Don't retry if the error was due to too many redirects.
+			if redirectsErrorRe.MatchString(v.Error()) {
+				return false, v
+			}
+
+			// Don't retry if the error was due to an invalid protocol scheme.
+			if schemeErrorRe.MatchString(v.Error()) {
+				return false, v
+			}
+
+			// Don't retry if the error was due to TLS cert verification failure.
+			if _, ok := v.Err.(x509.UnknownAuthorityError); ok {
+				return false, v
+			}
+		}
+
+		// The error is likely recoverable so retry.
+		return true, nil
+	}
+
+	// Check the response code. We retry on 500-range responses to allow
+	// the server time to recover, as 500's are typically not permanent
+	// errors and may relate to outages on the server side. This will catch
+	// invalid response codes as well, like 0 and 999.
+	if resp.StatusCode == 0 || (resp.StatusCode >= 500 && resp.StatusCode != 501) {
+		return true, fmt.Errorf("unexpected HTTP status %s", resp.Status)
+	}
+
+	return false, nil
+}
+
 // DefaultBackoff provides a default callback for Client.Backoff which
 // will perform exponential backoff based on the attempt number and limited
 // by the provided minimum and maximum durations.
@@ -509,9 +551,13 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 	}
 
 	var resp *http.Response
-	var err error
+	var attempt int
+	var shouldRetry bool
+	var doErr, checkErr error
 
 	for i := 0; ; i++ {
+		attempt++
+
 		var code int // HTTP response code
 
 		// Always rewind the request body when non-nil.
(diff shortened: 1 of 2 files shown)


theckman:
@evanphx I think this is ready for review if you're interested / when you have a moment.

client.go
```diff
@@ -357,7 +357,7 @@ func DefaultRetryPolicy(ctx context.Context, resp *http.Response, err error) (bo
 	// errors and may relate to outages on the server side. This will catch
 	// invalid response codes as well, like 0 and 999.
 	if resp.StatusCode == 0 || (resp.StatusCode >= 500 && resp.StatusCode != 501) {
-		return true, nil
+		return true, fmt.Errorf("unexpected HTTP status %s", resp.Status)
```
evanphx: This is likely a backwards compatibility breakage that we should not do.

client.go
```diff
@@ -542,19 +545,26 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		}
 	}
 
-	if c.ErrorHandler != nil {
+	defer func() {
```
evanphx: Pretty sure this is wrong as it will close all Idle Connections on every return, which means no keepalive will be used. In the non-error case below on lines 552, we keep them alive between successful requests.

client.go
theckman: I took this as a less-severe breaking change, because I don't see a way to accomplish better propagation of errors without breaking the API.

Is there any suggestion you can make around how to accomplish this without the change? What's the evaluation criteria for a breaking change when it adds valuable functionality?

client.go
theckman: Wasn't my intent, let me look at this area again.

client.go
theckman: @evanphx I guess looking at this comment further, whether it's a breaking change might be up to who you ask. I've seen similar changes make it in to Go as a non-breaking change (since code still compiles), but the argument could definitely be made that it's breaking to some degree.

client.go
```diff
@@ -542,19 +545,26 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
...
 	// returning the response
 	if resp != nil {
-		resp.Body.Close()
+		c.drainBody(resp.Body)
 	}
-	c.HTTPClient.CloseIdleConnections()
-	return nil, fmt.Errorf("%s %s giving up after %d attempts",
-		req.Method, req.URL, c.RetryMax+1)
+
+	return nil, fmt.Errorf("%s %s giving up after %d attempt(s): %v",
```
ryanuber: This seems like a good candidate to change to a `%w` - that way the underlying error can be unwrapped if desired. https://blog.golang.org/go1.13-errors

jefferai: Is `%w` compatible with 1.12?

ryanuber: Hm, perhaps not. Is that an issue? Not a huge deal and we can punt on doing %w for now, but it would be nice to get the additional error if possible.

jefferai: Not sure if it's an issue. May just want to call out in readme the go version required.

tgropper:
Wouldn't it be better if we change the returned error to a typed model (instead of `errors.errorString`) and wrap the original request error?

Also it would be nice that if there is a non-successful status code we add to the error the response body, considering the server may return a more detailed error description.

client.go
theckman: Sorry that this took so long to wrap back around on. I'm going to be looking to address this, and the other feedback shortly.

client.go
theckman: We can declare it in the `go.mod` file at a minimum. We could also prevent people from building this package on < go1.13 as well. That might be a bit heavy-handed.

theckman: What are your thoughts on that?

client.go
theckman: @evanphx I can revert this change back so as not to break your API. Would you see value in offering this as a retry policy function from the package, but under a different name?

theckman:
@tgropper would you be able to clarify what you mean by a "typed model"? I'm not familiar with that term.

To your second point, could you clarify that a bit more? Are you suggesting we read the response body and append it to the error value we return?

client.go
theckman: Oh, nevermind. Looks like it's already there. So maybe nothing to do here, and we can make use of `%w`.

client.go
theckman: I'm now realizing after a rebase that I need to also address the new error handling introduced above. I'll create an retry policy function with error propagation, and undo my breaking changes to this one.

theckman: @evanphx I've moved this out to its own function now, so it's no longer a breaking change.

theckman:
@evanphx this is ready for review, if you're still willing/able.

client.go
```diff
@@ -432,6 +432,48 @@ func DefaultRetryPolicy(ctx context.Context, resp *http.Response, err error) (bo
 	return false, nil
 }
 
+// ErrorPropagatedRetryPolicy is the same as DefaultRetryPolicy, except it
+// propagates errors back instead of returning nil. This allows you to inspect
+// why it decided to retry or not.
+func ErrorPropagatedRetryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	// do not retry on context.Canceled or context.DeadlineExceeded
+	if ctx.Err() != nil {
```
evanphx, line 440: I think this should use the non-blocking check on ctx here instead, ie: ```
select {
case <-ctx.Done():
   return false, ctx.Err()
default:
  // ok, keep going
}
```

Comment (theckman) [reply client.go:440]: @evanphx I can change it if you'd prefer it. The contract as specified in the `context.Context` interface indicates it's not necessary:

```go
// If Done is not yet closed, Err returns nil.
// If Done is closed, Err returns a non-nil error explaining why:
// Canceled if the context was canceled
// or DeadlineExceeded if the context's deadline passed.
// After Err returns a non-nil error, successive calls to Err return the same error.
```

My understanding is that the channel mechanism is for when you want to use a `select` between multiple channels, one being the context's `Done()` channel. 

Please let me know if you want the change before merging.

evanphx: Ah! I had forgotten the docs said that. No problem, leave as is.

jefferai:
LGTM too, merging! Thanks!

PR #72, opened 2019-10-25
Add CircleCI & remove travis

mdeggies:
This PR moves from travis to CircleCI, using gotestsum for testing. Setup is very similar to those in other repos, see https://github.com/hashicorp/go-version as an example.

Tests pass locally using the latest go version, 1.13.3:

tests


kpenfound:
Awesome! Don't forget to update the build status indicator in `README.md`

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
+version: 2.1
+
+references:
+  images:
+    go: &GOLANG_IMAGE circleci/golang:latest
```
alvin-huang: For the products, we normally will pin it to a specific version so we don't have to dig through and compare when builds break to find out that `latest` was a version bump that caused issues. In this case if it works for the latest one, you can pick that one and set it and we can just manually bump it when needed in the future.

alvin-huang:
One small nit pick but LGTM!

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+      - TEST_RESULTS: *TEST_RESULTS_PATH
+
+jobs:
+  go-test:
+    executor: go
+    steps:
+      - checkout
+      - run: mkdir -p $TEST_RESULTS
+
+      - restore_cache: # restore cache from dev-build job
```
RussellRollins: Uber-nit: `dev-build` doesn't seem to be a job in this config, this comment is confusing.

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+      # check go fmt output because it does not report non-zero when there are fmt changes
+      - run:
+          name: check go fmt
+          command: |
+            files=$(go fmt ./...)
+            if [ -n "$files" ]; then
+              echo "The following file(s) do not conform to go fmt:"
+              echo "$files"
+              exit 1
+            fi
```
samsalisbury: I like it, I would have been tempted to use `if [ -n "$(gofmt -d ./...)" ]; then ...` but I like yours better as it lists the file names at fault.

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+      - run: |
+          PACKAGE_NAMES=$(go list ./...)
+          gotestsum --format=short-verbose --junitfile $TEST_RESULTS/gotestsum-report.xml -- $PACKAGE_NAMES
+      - store_test_results:
+          path: *TEST_RESULTS_PATH
+      - store_artifacts:
+          path: *TEST_RESULTS_PATH
+
+workflows:
+  version: 2
```
samsalisbury: I believe that under 2.1 syntax, you can drop this `version: 2` line in workflows.

.circleci/config.yml
mdeggies: Thanks! QQ: I'm setting up the project in circle: https://circleci.hashicorp.engineering/setup-project/gh/hashicorp/go-retryablehttp and by default, it builds on every branch. So after the setup and a commit, it should do a fresh build. Any other steps I'm missing?

.circleci/config.yml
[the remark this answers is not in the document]
samsalisbury: Also, echoing @esacteksab in Slack, usually a good idea to pin to a specific Go version.

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+      - save_cache:
+          key: go-retryablehttp-modcache-v1-{{ checksum "go.mod" }}
+          paths:
+            - "/go/pkg/mod"
+
+      # check go fmt output because it does not report non-zero when there are fmt changes
+      - run:
+          name: check go fmt
+          command: |
+            files=$(go fmt ./...)
```
samsalisbury: Be careful here, as this will fail if there is a syntax error in the go code, but will place that error on stderr so you'll miss it in the condition below. You could either just redirect stderr to capture it (`files=$(go fmt ./... 2>&1)`) or switch on automatic error detection in bash (see above).

.circleci/config.yml
samsalisbury: @mdeggies for some reason, CircleCI won't build anything until there is some valid CircleCI config on master, so I usually just push a very minimal config to master without a PR just to get it building first.

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+              exit 1
+            fi
+      # run go tests with gotestsum
+      - run: |
+          PACKAGE_NAMES=$(go list ./...)
+          gotestsum --format=short-verbose --junitfile $TEST_RESULTS/gotestsum-report.xml -- $PACKAGE_NAMES
+      - store_test_results:
+          path: *TEST_RESULTS_PATH
+      - store_artifacts:
+          path: *TEST_RESULTS_PATH
```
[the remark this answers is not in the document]
mdeggies: Resolving, since I removed all anchors

.circleci/config.yml
mdeggies: Thanks! Removed

.circleci/config.yml
mdeggies: Thanks guys.

mdeggies:
> Hi @mdeggies, this looks great, comments below. The only one that is semi-blocking is handling the case that `go fmt` fails due to syntax error, other than that I'm just being extra picky and verbose to pass on some of the lessons I've learnt so far writing CircleCI config, feel free to use or ignore whatever you see fit.

Thanks for all the notes @samsalisbury! I added the project on CircleCI, and tweaked the settings to make it build on PR's from forks. Also removed all anchors and converted them to commands. `get_test_results_path` just returns the path, and `store_results` actually stores the artifact and test results. Let me know if this looks good- if so, I'll add the same logic to the other PR's. 

Latest build: https://circleci.com/gh/hashicorp/go-retryablehttp/8

.circleci/config.yml
```diff
@@ -1,12 +1,67 @@
...
     docker:
       - image: circleci/golang:1.13.3
+    shell: /usr/bin/env bash -euo pipefail -c
+
+jobs:
+  go-test:
+    executor: go
     steps:
+      - checkout
+      - run: mkdir -p get_test_results_path
```
samsalisbury: Sorry, I think my comment about this was a bit unclear. You can only reference commands as `steps`, they don't get expanded in places like this, so this just results in the literal string `get_test_results_path`.

.circleci/config.yml
```diff
@@ -1,12 +1,67 @@
...
-jobs:
-  build:
+commands:
+  get_test_results_path:
+    parameters:
+      path:
+        type: string
+        default: /tmp/test-results
+    steps:
+      - run: echo << parameters.path >>
```
samsalisbury: I think it's not possible to use this in the way it's intended. When we `run` a command, it happens in isolation in its own shell, so we can't dynamically read the output from executed commands in this way. Commands themselves are are used in the same way the `run` step is used, and can themselves contain multiple `run` steps, or references to other commands in their place.

theckman:
Looks like this needs to be upped to the latest version of Go (`1.14.2`).

mdeggies:
> Looks like this needs to be upped to the latest version of Go (`1.14.2`).

Thanks for the ping on this.. I'll close in favor of a fresh PR.

PR #73, opened 2019-10-25
Update default retry policy to fast-path likely unrecoverable errors

ryanuber:
This PR addressed two cases I believe our default retry policy should not retry:

* Unverifiable TLS certificate (self-signed cert) - it's not likely that this error is going to be recoverable within the timeframe of the retries.

* Too many redirects - The path the client gets redirected through is unlikely to change significantly enough to warrant a retry.

I considered making a new policy for these, but it seemed like widely desirable behavior and I don't think this is likely going to break anything, so I went for the default policy.

### DIFF
--- a/client.go
+++ b/client.go
@@ -34,6 +34,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


jefferai:
Two things:

1) Did you go mod tidy? The go.sum changes seem more verbose than they need to be. Not a huge deal though.

2) I have actually seen this in the wild where an LB points to many working servers and one misconfigured server. I think it's a good change, just pointing out that it's possible cause I've seen it :-)

ryanuber:
Thanks!

1. Forgot, pushed!
2. It's a fair point. I was just going for what I think is probably the majority cases for the default policy. If you think we should have a separate policy for these tweaks, I'm fine doing that also. This change is just based on personal experience waiting for retries to complete when I know they are never going to.

evanphx:
@ryanuber Not to complicate the solution any further, but I wonder if the ability to short the wait and try again would be useful here. Namely, if you hit one of these errors, immediately try again rather than waiting and if that fails, then just give up. Users can obviously do that themselves by providing their own version of the function.

ryanuber:
@evanphx yeah I think that would be a more cross-cutting change because it `CheckRetry` by itself can't influence the backoff (which is configured separately by the customizable `Backoff` function). Interesting idea, but definitely more than I was planning to tackle for this case.

jefferai:
LGTM!

PR #81, opened 2020-02-14
Add TotalWaitMax to cap the total time for issuing request and retries

pmalekn:


pmalekn:
@jefferai Sure, so how about 3 minutes then?

jefferai:
Sounds good to me.

jefferai:
@ryanuber You good with this feature?

ryanuber:
Before I comment on this feature, @pmalekn had you considered using a `context` with a timeout to achieve the same? Something like:

```go
ctx = context.WithTimeout(ctx, 3*time.Minute)
req := retryablehttp.NewRequest("GET", "https://foo.bar", nil).WithContext(ctx)
```

One benefit to note here is that this would have the ability to interrupt `net/http` in the midst of a request, keeping the actual time spent closer to the value you desire.

pmalekn:
@ryanuber I thought about this but it occurred to me that it's too verbose
Nonetheless I agree that it's better in the sense that it can interrupt `net/http` in the middle of making the request.

How about using `context.WithTimeout()` and `.WithContext` inside `func (c *Client) Do(req *Request) (*http.Response, error)`?

pmalekn:
The slight problem that we might run into is that `.WithContext()` creates a "shallow copy" of provided request (which we probably would like to avoid).

jefferai:
I forgot we added WithContext! Ryan is correct, that's a better idea.

I don't think there is anything wrong with a shallow copy.

pmalekn:
@jefferai @ryanuber Do you mind taking a look? I'd refactored to use `.WithContext()`

jefferai:
I think Ryan's point was that this is unnecessary altogether, since you can already use the context to have this functionality.

ryanuber:
@jefferai is correct, I was suggesting that you could incorporate `WithContext()` into your use of go-retryablehttp from the caller side in order to get the functionality you're after.

I can appreciate that what you want is a client-level setting for the maximum overall time of a request. Providing an appropriate `context` is a good way to go about doing this on a per-request basis. Alternatively, tuning the client-level values of `RetryWaitMin`, `RetryWaitMax`, `RetryMax`, and `Backoff` should give you very predictable results, less the actual time spent in the network layer.

jefferai:
Closing for now based on feedback.

PR #84, opened 2020-02-18
Added wrapper for compatibility with http.Client

### LINKED ISSUE #77: http.Client compatible NewHTTPClient?
Hello! I like your functionality, but I'm in a situation where I need to adhere to the http.Client api exactly, and the `Do()` signature is not `func (c *Client) Do(req *http.Request) (*http.Response, error)`.

Lots of enhanced/replacement libraries seem to implement the old api exactly so you can just swap out an import (pkg/errors vs. stdlib errors for instance). As a workaround, a wrapper is possible:
```
type RetryingHTTPClient struct {
    client *retryablehttp.Client
}

func (c *RetryingHTTPClient) Do(req *http.Request) (*http.Response, error) {
    wrappedReq, err := retryablehttp.FromRequest(req)
    if err != nil {
        return nil, err
    }

    return c.client.Do(wrappedReq)
}
```
Is there any openness to a PR adding a method like `NewHTTPClient()` that would return a client that conformed to something like `func (c *RetryingHTTPClient) Do(req *http.Request) (*http.Response, error)`? 
```
// HTTPDoer is an interface for the one method of http.Client that is used by RetryingHTTPClient
type HTTPDoer interface {
	Do(*http.Request) (*http.Response, error)
}
```

xonstone:
Fixes #77


jefferai:
Honestly, I'm kind of torn between this and the suggestion at https://github.com/hashicorp/go-retryablehttp/issues/77#issuecomment-600939673

Making it a RoundTripper adds this functionality to anything using RoundTripper, including a custom stdlib http.Client.

xonstone:
I feel like the RoundTripper might be the wrong abstraction for this. This for 3 reasons:

1.  The documentation of [RoundTripper ](https://golang.org/pkg/net/http/#RoundTripper) clearly states that it should execute a single http request.
2. The rest of retryablehttp already follows the interface of http.Client closely.
3. The other implementations of RoundTripper in stdlib seem to be dealing with lower-level operations e.g. [NewFileTransport](https://golang.org/src/net/http/filetransport.go?s=827:876#L20)

jefferai:
> The documentation of RoundTripper clearly states that it should execute a single http request.

Not quite:

> RoundTripper is an interface representing the ability to execute a single HTTP transaction, obtaining the Response for a given Request. 

The output from the client perspective is the Response for the Request; the fact that under-the-hood this library may have performed some retries in order to actually get that Response is an implementation detail.

Keep in mind both possibilities return an http.Client; the RoundTripper version is just more minimal. What's in this PR embeds a retryablehttp.Client so if you want to modify the http.Client you can but you're reaching in two levels deep. But it's there. In the RoundTripper version you track the client independently.

Probably that means that this is a better idea if for no other reason that you don't need to track the retryable Client separately from the http.Client if you want to make modifications. Unsure though. @ryanuber thoughts?

xonstone:
Still feels like it's violating the interface of RoundTripper

Pulled from the docs:

```
    // RoundTrip should not attempt to interpret the response. In
    // particular, RoundTrip must return err == nil if it obtained
    // a response, regardless of the response's HTTP status code.
    // A non-nil err should be reserved for failure to obtain a
    // response. Similarly, RoundTrip should not attempt to
    // handle higher-level protocol details such as redirects,
    // authentication, or cookies.
```

ryanuber:
I'm a bit torn as well. One thing that stands out to me is that in many cases, you're probably dealing with a struct or a function which directly expects an `*http.Client`, in which case you would not be able to pass in an instance of this shim without also creating yet another interface for the `Do()` method. With the `RoundTripper` method, you can easily slide a `*retryablehttp.Client` underneath any normal `*http.Client` and pass that around freely.

I do agree though that this doesn't feel like a perfect fit for the `RoundTripper` interface.

I think I'm ok with this approach, but the struct and method naming creates some confusion for me. The `New...()` method should be named exactly like the struct type it returns, so in this case that would be `NewRetryingHTTPClient`. However, to me it's not clear what the difference is from the `NewClient()` method, because technically they both return retrying HTTP clients. I am lacking creativity at the moment myself, so my only real suggestions would be `StandardClient` or `CompatClient`. Open to other ideas though.

jefferai:
I like StandardClient as it matches StandardLogger from https://godoc.org/github.com/hashicorp/go-hclog#Logger

@ryanuber you bring up some good points about RoundTripper. At this point I'm thinking that there are reasons to offer _both_ approaches. e.g. StandardClient and StandardRoundTripper. That way instead of us trying to guess at what a particular situation might call for, it's available either way.

xonstone:
I've renamed it to StandardClient

standard_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func TestStandardClient_Do(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "POST" {
+			t.Fatalf("bad method: %s", r.Method)
+		}
+
+		bytes, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			t.Fatalf("bad body: %s", r.Body)
```
ryanuber, line 20: You want to dump the `err` here because there was a problem reading it.

standard_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestStandardClient_Do(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "POST" {
```
ryanuber, line 14: You've used `http.MethodPost` in a few other places, would be nice to shore this up with that.

standard_test.go
```diff
@@ -0,0 +1,85 @@
...
+			t.Fatalf("bad method: %s", r.Method)
```
ryanuber, line 15: It does technically fail the test to call `t.Fatalf` within the handler, but what's odd is that the failures will appear outside of the actual test context (other than an EOF error on the client), and will halt the remainder of the tests from running since this is out in the main context. Not a huge deal, and I'd even accept this as-is, but I would have probably had the HTTP server respond with an appropriate status code instead, and checked that from the test case.

standard_test.go
```diff
@@ -0,0 +1,85 @@
...
+			t.Fatalf("bad body: %s", r.Body)
+		}
+
+		w.WriteHeader(200)
+	}))
+	defer server.Close()
+
+	type args struct {
+		req *http.Request
+	}
```
ryanuber, line 33: This struct seems unnecessary. I'd just have `req` directly in the test case struct.

standard_test.go
```diff
@@ -0,0 +1,85 @@
...
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
```
ryanuber, line 37: One pattern I've found more useful in the past is to have this be a string such that you can put part of the expected error message in it. This way, you can't get a bad test result from say a failing request vs. a bad request.

standard_test.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			c := NewStandardClient()
+
+			_, err := c.Do(tt.args.req)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Do() error = %v, wantErr %v", err, tt.wantErr)
+				return
```
ryanuber, line 75: I think you can just use `t.Fatalf` as you do elsewhere. No return necessary.

standard.go
```diff
@@ -0,0 +1,23 @@
+package retryablehttp
+
+import "net/http"
+
+// NewStandardClient creates a new StandardClient with a Client with default settings.
+func NewStandardClient() StandardClient {
```
ryanuber, line 6: I think this should be returning a `*StandardClient` considering that's what the `Do()` method operates on.

xonstone:
Thanks for the feedback! Adjusted accordingly.

jefferai:
Will review again after RoundTripper support is added.

ryanuber:
@xonstone were you planning to add `RoundTripper` support yourself? If so we can keep this open. Otherwise, we can probably accept this PR in isolation, and add `RoundTripper` in a separate PR. @jefferai any reason you wanted both in this PR? I could be missing something.

xonstone:
I was not planning on adding this...

jefferai:
Sure, I wanted it in this PR because it doesn't seem like this functionality works for all use-cases. It seemed if we were going to have http.Client support we might as well have it usable for all needs.

ryanuber:
After thinking about this a bit, I think the most ideal setup would be this:

`func (c *Client) StandardClient() *http.Client` - Returns a stdlib `*http.Client` with the transport underneath set up to use a customer `RoundTripper`, which leverages retryablehttp. This would match the `go-hclog` functionality perfectly, and also gives us a really easy way to setup the retryablehttp client options and then get a stdlib http client that uses it.

I do still see the merit in this PR's changes. My only hesitation at this point is the public API it introduces, which would no longer make sense after the above is introduced, and could cause some naming confusion. I think we could deal with this pretty easily though. Just wanted to highlight the possibility of a future deprecation of this specific implementation.

For now, this makes sense and is really easy. Thanks for tackling it @xonstone !

jefferai:
@ryanuber I think what you suggest sounds good, and a nice thing about it is that you could also just pull the underlying RoundTripper from the http.Client if you want to use it directly.

It's not clear to me why you suggested a preference for something else and then merged this :-) Why not revert this, and do that instead? Then we don't need to worry about deprecations.

PR #86, opened 2020-03-01
example usage with jitter

KlotzAndrew:
currently there are not too many docs for someone to quickly get started

added a small example for wrapping a POST request with jitter, a use
case I immediately had


theckman:
@KlotzAndrew are you able to sign the CLA so this PR has all green marks on it? Might help it get merged.

PR #88, opened 2020-03-31
Add SetBody and WriteTo methods to Request

ash2k:
1. Adds `SetBody()` method to allow mutating the body not via `.Body` but with a method that also sets content length and properly updates the private field.
2. Adds `WriteTo()` method to efficiently stream the request body into an `io.Writer`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -119,6 +119,19 @@ func (r *Request) BodyBytes() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// SetBody allows setting the request body.
+//
+// It is useful if a new body needs to be set without constructing a new Request.
+func (r *Request) SetBody(rawBody interface{}) error {
+	bodyReader, contentLength, err := getBodyReaderAndContentLength(rawBody)
+	if err != nil {
+		return err
+	}
+	r.body = bodyReader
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -119,95 +119,110 @@ func (r *Request) BodyBytes() ([]byte, error) {
...
+	r.body = bodyReader
+	r.ContentLength = contentLength
+	return nil
+}
+
 func getBodyReaderAndContentLength(rawBody interface{}) (ReaderFunc, int64, error) {
 	var bodyReader ReaderFunc
 	var contentLength int64
 
-	if rawBody != nil {
```
ash2k: I've replaced this `if` check with a `case` statement below. Please tick `Hide whitespace changes` above to see that there are no other changes.

jefferai:
I notice that you did some updating a few hours ago -- let me know if this is actually ready to merge.

Also, any reason not to put the nil check in the switch statement the first case?

ash2k:
@jefferai Thanks for the quick review! This is ready to merge if you think it's good. I can move the `nil` check up if you prefer, it does not matter where it is.

jefferai:
Cool, thanks again.

PR #90, opened 2020-04-13
Add RoundTripper and updated StandardClient to leverage it

ryanuber:
Adds an `http.RoundTripper` implementation, which actually uses retryablehttp under the covers. This should make retryablehttp usable more easily in situations where you need to directly use a `*http.Client`, and it simplifies the `StandardClient` enough that we don't even need the shim struct for it any longer.


ryanuber:
cc @xonstone - Wanted to ping you because we just merged #84 earlier today, but upon some further discussion (and a little spare time to implement this), we thought this was a better route to go, and it should satisfy your use case as well.

xonstone:
It indeed fits my use-case as well. Thanks for mentioning it!

bdimcheff:
👏 thanks!

PR #92, opened 2020-04-23
Prevent race during Client.HTTPClient initialization.

### LINKED ISSUE #89: Client.Do data race on Client.HTTPClient
When `Client.HTTPClient` is `nil` and there are concurrent calls to `Client.Do`, a data race can occur at: https://github.com/hashicorp/go-retryablehttp/blob/27fc49b35882e89429a3cf613f70e1c580bcd8a9/client.go#L493-L495

This can be resolved by doing the check and initialization inside a `sync.Once.Do`. I can open a PR if you'd like.

vcabbage:
Fixes #89

### DIFF
--- a/client.go
+++ b/client.go
@@ -357,6 +357,7 @@ type Client struct {
 	ErrorHandler ErrorHandler
 
 	loggerInit sync.Once
+	clientInit sync.Once
 }
 
 // NewClient creates a new Client with default settings.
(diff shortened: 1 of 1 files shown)


jefferai:
Thanks!

likexian:
@jefferai 
Same issue, When will this pr be released?

PR #95, opened 2020-05-19
Errors from original http.Client response are double-wrapped when using the RoundTripper

andrewlouis93:
When you're using a `http.Client` with a custom `RoundTripper` that actually facilitates the retries  (it maintains its own internal `http.Client` object); any errors returned from the original `http.Client` are re-wrapped.

For instance, trying to do a Get request that returns an error response, and using the PassthroughErrorHandler to return the real failure would result in an error like this: 

```
Get "http://this-url-does-not-exist.com/": Get "http://this-url-does-not-exist.com/": dial tcp: lookup this-url-does-not-exist.com: no such host
```

This is basically an `*url.Error` object nested within another `*url.Error`'s `Err` field. 

We can just `.Unwrap()` it to make sure that we only get what we need. Open to other suggestions though.

### DIFF
--- a/roundtripper.go
+++ b/roundtripper.go
@@ -3,6 +3,7 @@ package retryablehttp
 import (
 	"errors"
 	"net/http"
+	"net/url"
 	"sync"
 )
 
@@ -41,7 +42,9 @@ func (rt *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	// Execute the request.
 	resp, err := rt.Client.Do(retryableReq)
-	if errors.Unwrap(err) != nil {
+	// If we got an error returned by standard library's `Do` method, unwrap it
+	// otherwise we will wind up erroneously re-nesting the error.
+	if _, ok := err.(*url.Error); ok {
 		return resp, errors.Unwrap(err)
 	}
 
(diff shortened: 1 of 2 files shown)


andrewlouis93:
cc @jefferai - Just wanted to ping a maintainer for some 👀 here.

andrewlouis93:
cc @ryanuber - in case you are able to help here :)

jefferai:
@andrewlouis93 at the moment no reviews can proceed since you have not signed the CLA.

andrewlouis93:
@jefferai Done, had committed this with a different email attached

jefferai:
I think this looks good -- do you know if this is only an issue with `*url.Error`? Basically I'm wondering if it's worth asserting before unilaterally unwrapping or if this will always be problematic.

andrewlouis93:
@jefferai That's a good consideration. In this case yes, [but only because `Do` in the standard library only ever returns `*url.Error](https://github.com/golang/go/blob/2d77d3330537e11a0d9a233ba5f4facf262e9d8c/src/net/http/client.go#L582-L584); and the standard library `Do` invocations wind up getting nested under each other. 

We can be a bit more strict here and only unwrap if the error is an `*url.Error`; so we do not nest it again. 

Just updated and left some comments, take a look and let me know what you think.

jefferai:
LGTM, thanks!

PR #97, opened 2020-06-01
LeveledLogger: Add doc and prioritize it over Logger

ppai-plivo:
* Document LeveledLogger better to make it clear to developers that the
  second variadic argument accepts key-value pairs like hci-log or zap
  logger and not Printf style logging.
* When the passed logger implements both Logger and LeveledLogger
  interfaces, prioritize LeveledLogger over Logger. Switch statements
  evaluate cases from top to bottom.

### DIFF
--- a/client.go
+++ b/client.go
@@ -39,7 +39,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/hashicorp/go-cleanhttp"
+	cleanhttp "github.com/hashicorp/go-cleanhttp"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


ppai-plivo:
cc @jefferai for some 👀   on this. Thank you!

client.go
```diff
@@ -39,7 +39,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/hashicorp/go-cleanhttp"
+	cleanhttp "github.com/hashicorp/go-cleanhttp"
```
jefferai, line 42: This is an unnecessary change.

client.go
```diff
@@ -501,10 +505,10 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 
 	if logger != nil {
 		switch v := logger.(type) {
-		case Logger:
```
jefferai, line 504: I'm not sure that this was an accident. It might have been prioritized to keep behavior the same for backwards compat. @evanphx would know more.

evanphx: Looks like he just wanted to prioritize the LevelLogger over the generic one. That seems like a good change since LevelLogger is more specific.

client.go
ppai-plivo: It is unrelated but is intentional as it fixes the auto-removal of this import by goimports because in this case, the import path does not match the package name. Ref: https://github.com/golang/go/issues/28428

This slightly improves readability with no functional repercussions. I will remove this if you insist. Thank you.

cc @jefferai @evanphx

jefferai: So that's strange because imports prefixed by `go-` are auto-added just fine for me. That's why I suggested this wasn't necessary. I actually just tried taking a file and removing the `cleanhttp` and running `gomiports` on the file and it leaves it alone. Is it possible you're using a very old version of goimports or that you're not using go modules? IIRC this started working properly right around when go modules dropped on the scene.

client.go
jefferai: Sounds good then.

jefferai:
Thanks!

client.go
ppai-plivo: Sorry, my bad. You're right. I was using go modules but I had an older version of goimports.

PR #110, opened 2020-09-17
Replace http.Client with interface

asokol123:
### Changes
Created interface `HTTPClient` with functions `Do` and `CloseIdleConnections`
Replaced `HTTPClient` field type from `*http.Client` to `HTTPClient`
### Reason
This allows using custom clients.
We have custom http client with additional features, and we want to use it with retries. This module is awesome but now it's impossible to use it with out client.
P.S. Sorry for second PR, in first bot did't get that commit is mine.

### DIFF
--- a/client.go
+++ b/client.go
@@ -71,6 +71,11 @@ var (
 	schemeErrorRe = regexp.MustCompile(`unsupported protocol scheme`)
 )
 
+type HTTPClient interface {
+	CloseIdleConnections()
+	Do(req *http.Request) (*http.Response, error)
+}
+
 // ReaderFunc is the type of function that can be given natively to NewRequest
 type ReaderFunc func() (io.Reader, error)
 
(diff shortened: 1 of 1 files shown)


ryanuber:
Hey @asokol123 , thanks for the PR, and sorry for the delay. I'm hesitant to make a change like this, primarily because it is a breaking change (everyone will have to type cast before making direct modifications to the HTTPclient). retryablehttp has always been built directly over the top of `net/http`, and deviating from that isn't something that we're ready to change at the moment.

One thing I'll recommend is looking at net/http's `RoundTripper` interface and the `Transport` implementation of it. There's a lot you can do within the confines of the `RoundTripper` interface, and so it is possible that instead of a wholly custom client, you could implement a `RoundTripper` that would give you the same functionality underneath a standard `net/http` client.

Hope that helps. I'm going to close this for now, but if you have more questions please feel free to comment further.

PR #122, opened 2021-02-07
Set GetBody property when body is not empty (#121)

### LINKED ISSUE #121: Request wrapper make standard request `GetBody` property always nil
In the Go standard library, when creating a new request, the GetBody property of the request is set according to the body parameter.

This attribute will affect the client's behavior after receiving the 307, 308 response code. If it is empty, it is likely that it will not follow redirect.
```go
if ireq.GetBody == nil && ireq.outgoingLength() != 0 {
	// We had a request body, and 307/308 require
	// re-sending it, but GetBody is not defined. So just
	// return this response to the user instead of an
	// error, like we did in Go 1.7 and earlier.
	shouldRedirect = false
}
```
But now when request wrapper creates a new request, the body parameter is not passed, making GetBody always nil.
```go
func NewRequest(method, url string, rawBody interface{}) (*Request, error) {
	bodyReader, contentLength, err := getBodyReaderAndContentLength(rawBody)
	if err != nil {
		return nil, err
	}

	httpReq, err := http.NewRequest(method, url, nil)  // <=== here, body is nil
	if err != nil {
		return nil, err
	}
	httpReq.ContentLength = contentLength

	return &Request{bodyReader, httpReq}, nil
}
```

rgb-24bit:
When setting the body attribute of request wrapper, set the GetBody attribute of http request at the same time. (Fix #121)


jordans-openai:
Hi, we are facing similar issues. Any update on merging this fix?

timbunce:
Could this be closed now that #207 has been merged?

PR #126, opened 2021-02-27
client: support NewRequestWithContext

### LINKED ISSUE #125: Support NewRequestWithContext
Go added http.NewRequestWithContext in go 1.13 which fixes several of the issues with the original http.NewRequest in terms of instantiating the `*http.Request` with saner defaults and avoiding the need to copy if you want to set a *context.Context on the request.

It would be nice if this library supported that behavior.

kevinburkemeter:
See the docstring for http.NewRequestWithContext for a description of
the changes between http.NewRequest and http.NewRequestWithContext.

Fix a test that attempts to call Fatalf from a non-main goroutine,
which is not allowed in the upcoming Go 1.17 release.

Fixes #125.

### DIFF
--- a/client.go
+++ b/client.go
@@ -255,20 +255,25 @@ func FromRequest(r *http.Request) (*Request, error) {
 	return &Request{bodyReader, r}, nil
 }
 
-// NewRequest creates a new wrapped request.
-func NewRequest(method, url string, rawBody interface{}) (*Request, error) {
+// NewRequestWithContext creates a new wrapped request.
+func NewRequestWithContext(ctx context.Context, method, url string, rawBody interface{}) (*Request, error) {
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -255,14 +255,14 @@ func FromRequest(r *http.Request) (*Request, error) {
 	return &Request{bodyReader, r}, nil
 }
 
-// NewRequest creates a new wrapped request.
-func NewRequest(method, url string, rawBody interface{}) (*Request, error) {
+// NewRequestWithContext creates a new wrapped request.
```
ryanuber, line 258: I would go a little more descriptive here and mention what happens with the context.

ryanuber (commented):
This looks good. I think you could write a super easy test just to ensure the context actually gets passed appropriately. If you could flesh out the comment that I mentioned, and add a small test for this, I think this can be merged.

PR #128, opened 2021-03-10
Add support for CheckRedirect in Client.StandardClient()

nuttmeister:
Hello everyone!

This PR adds support for custom redirect behavior like the standard http.Client supports with the `CheckRedirect` function.
Before this behavior couldn't be changed, not even on the default client returned by `StandardClient()`.

It works using both `*Client` and `*http.Client` from `StandardClient()`.
For example, disabling follow redirect all together as follows.

```go
    client := retryablehttp.NewClient()
    client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
        return http.ErrUseLastResponse
    }
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -769,6 +769,7 @@ func (c *Client) PostForm(url string, data url.Values) (*http.Response, error) {
 // shims in a *retryablehttp.Client for added retries.
 func (c *Client) StandardClient() *http.Client {
 	return &http.Client{
-		Transport: &RoundTripper{Client: c},
+		Transport:     &RoundTripper{Client: c},
+		CheckRedirect: c.HTTPClient.CheckRedirect,
 	}
 }
--- a/client_test.go
+++ b/client_test.go
@@ -648,6 +648,68 @@ func TestClient_CheckRetryStop(t *testing.T) {
 	}
 }
 
+func TestClient_CheckRedirects(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Location", "/new/path")
+
+		switch r.URL.Path {
+		case "/301":
+			w.WriteHeader(301)
+		case "/302":
+			w.WriteHeader(302)
+		default:
+			w.WriteHeader(500)
+		}
+	}))
+	defer ts.Close()
+
+	client := NewClient()
+	client.HTTPClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		return http.ErrUseLastResponse
+	}
+	stdClient := client.StandardClient()
+
+	tests := []int{301, 302}
+
+	// Check that we get 301 and 302 responses.
+	for _, test := range tests {
+		resp, err := client.Get(fmt.Sprintf("%s/%d", ts.URL, test))
+		if err != nil {
+			t.Fatalf("unexpected error testing check redirect. %s", err.Error())
+		}
+		if resp.StatusCode != test {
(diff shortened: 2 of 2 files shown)


nuttmeister:
To clarify the check in Do could possibly be removed by using a setter function instead of exported variable.
But my thinking is since setters are not used for the other configuration it would lead to confusion.

So that's why the check is there so you could create the client, do a request, change the CheckRedirect variable and the next request would work.

Cheers!

ryanuber:
Hey @nuttmeister ! Thanks for opening this.

> Before this behavior couldn't be changed, not even on the default client returned by StandardClient().

Could you clarify why you are unable to set the `CheckRedirect` on the standard `*http.Client`? If you configure it on retryablehttp's `HTTPClient` field of the `*Client` struct, it seems like that ought to take care of it.

```go
c := retryablehttp.NewClient()
c.HTTPClient.CheckRedirect = ...
```

I could be missing something, let me know!

nuttmeister:
@ryanuber hey! I tried setting it on both `client.HTTPClient.CheckRedirect` and `client.StandardClient().CheckRediect`.
Worth nothing that in both cases I'm using the standard client after (by using `StandardClient()`).

So totally possible some of this PR is not really necessary. Didn't think that one through 100%.
I will check it out and do an update if necessary removing that part if that is the case.

nuttmeister:
@ryanuber I have narrowed it down basically (I think).
Sorry for not being more thorough on this first.

Using just the regular none standard `Client` it works by just setting `Client.HTTPClient.CheckRedirect`.

But if you use the standard `http.Client` by using `Client.StandardClient()` it will not work by first have set `Client.HTTPClient.CheckRedirect`. Neither does setting `http.Client.CheckRedirect` that is returned.

However, what DOES does work that I just found out is actually doing this:
```go
client := NewClient()
client.HTTPClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
    return http.ErrUseLastResponse
}
stdClient := client.StandardClient()
stdClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
    return http.ErrUseLastResponse
}
```

Which is not really straight forward perhaps to get it working when using the standard client.

But yeah, I will refactor the PR to copy these settings from what has been set in the `HTTPClient` instead.
Thanks for putting me on the correct path, haha.

nuttmeister:
@ryanuber work good with copying `HTTPClient.CheckRedirect` to the new client returned by `StandardClient()`.
Will update the PR.

That saying. I guess it would be good to also copy `Timeout` and the `Jar` as well? Or what do you think?
Since you most likely would like to have use those with the standard client if they where set on `Client.HTTPClient`.

nuttmeister:
Updated to just copy the the `CheckRedirect` from HTTPClient.

nuttmeister:
Long time with not activity. Is this something we want or should I close it?

jefferai:
The CLA isn't signed yet so we cannot do anything with this PR at the moment, apologies.

nuttmeister:
Hello. It was signed before. However I have changed companies twice and this email. The question is if it's wanted or not. If so I can try and sort it out.

jefferai:
Yes, I think it's a good idea.

manicminer:
@nuttmeister Wondering if you're able to get this ready to merge? Thanks!

nuttmeister:
@manicminer sure, I just totally forgot about this. I will have a look at it today.

nuttmeister:
@manicminer there, force pushed with the now correct email that has signed the agreement.

manicminer:
@nuttmeister Thanks, appreciate you picking up this old PR, I'm glad I can preserve your commits :)

PR #131, opened 2021-04-24
Add StatusServiceUnavailable to Retry-After check in DefaultBackoff

vancluever:
This adds StatusServiceUnavailable (error 503) to the codes checked for
the Retry-After header in DefaultBackoff.

This is a valid header to honor for 503 responses. See:
  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After


mwhooker (approved):
lgmt.

PR #133, opened 2021-05-12
logs: redact in URL passwords

drakkan:
So if we have URL like this one http://user:password@127.0.0.1/ the password is not logged


drakkan:
the `Redacted` method was added in [1.15](https://github.com/golang/go/issues/34855) so tests fail on previous Go versions.

PR #138, opened 2021-08-09
support HTTPDate in Retry-After header

justenwalker:
According to [RFC7231](https://httpwg.org/specs/rfc7231.html#header.retry-after)
the Retry-After header may be specified in both number of seconds OR an HTTPDate

>  The value of this field can be either an HTTP-date or a number of seconds to delay after the response is received.
>
>        Retry-After = HTTP-date / delay-seconds
>
>  A delay-seconds value is a non-negative decimal integer, representing time in seconds.
>
>        delay-seconds  = 1*DIGIT
>
>  Two examples of its use are
>
>       Retry-After: Fri, 31 Dec 1999 23:59:59 GMT
>       Retry-After: 120

This change adds support for parsing the date if the format of the header contains non-numeric characters

Closes: #144


Opelord:
Hi, I have seen a few pull requests like this that are updated whenever the last pull request have conflicts with current version. Can Hashicorp add this functionality, please?

justenwalker:
I opened #144 for discussion about the feature. Given that there are now 3 PRs that accomplish the same thing, we should probably consolidate discussion in a single place.

manicminer (approved):
@justenwalker Thanks for submitting this PR! Although there are a few open PRs aiming to add this functionality, this feels like the cleanest and most robust implementation. As such I'll merge this for inclusion in an upcoming release. Thanks again.

PR #140, opened 2021-09-06
WithContext should return a new Request

ash2k:
The method should not mutate the original object. That is how the method from stdlib works.


ryanuber:
Good catch, thanks for taking care of this 👍

PR #147, opened 2021-10-19
HTTP client factory for per-request clients

ash2k:
Please see https://github.com/hashicorp/go-retryablehttp/pull/139#discussion_r729367732 for the background.

This approach allows the library user to decide how they want to handle any client/transport related concerns. It decouples the library from those concerns as it doesn't have to have an opinion on how keep-alives need to work. Different situations need different behavior so giving users full control is the best option.

### DIFF
--- a/client.go
+++ b/client.go
@@ -401,11 +401,24 @@ type ErrorHandler func(resp *http.Response, err error, numTries int) (*http.Resp
 // PrepareRetry is called before retry operation. It can be used for example to re-sign the request
 type PrepareRetry func(req *http.Request) error
 
+type HTTPClient interface {
+	// Do performs an HTTP request and returns an HTTP response.
+	Do(*http.Request) (*http.Response, error)
+	// Done is called when the client is no longer needed.
+	Done()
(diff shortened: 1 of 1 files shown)


ash2k:
@ryanuber PTAL

client.go
```diff
@@ -779,3 +798,29 @@ func (c *Client) StandardClient() *http.Client {
...
+
+var (
+	_ HTTPClientFactory = &CleanPooledClientFactory{}
+	_ HTTPClient        = &idleConnectionsClosingClient{}
+)
+
+type CleanPooledClientFactory struct {
+}
+
+func (f *CleanPooledClientFactory) New() HTTPClient {
```
ash2k, line 810: Something to consider is adding an `error` to the returned values. Some factories may have a problem while constructing a client and they will not have a good way to propagate the error up. WDYT?

ash2k:
Rebased. This is ready for review/merge.

ash2k:
Rebased again. 3 years and counting :)

PR #149, opened 2021-11-09
Better comments for `err == nil` scenario

devosnw:
Lemme know if this make more sense. I don't think this comment was entirely accurate. 

I think we can only get to this block if we've had no errors (from `Do` and `CheckRetry`) but somehow we still wanted to retry.

### DIFF
--- a/client.go
+++ b/client.go
@@ -690,8 +690,8 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		c.drainBody(resp.Body)
 	}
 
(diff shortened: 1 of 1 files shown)


devosnw:
Gonna close this actually. I understand better what's going on here. I still don't love the original comments but mine aren't good either. `err == nil` is what will generally happen when using `DefaultRetryPolicy` and the below case is what will happen when using `ErrorPropagatedRetryPolicy`

PR #155, opened 2022-02-10
Change URL from godoc to pkg.go.dev in README.md

mkevac:
Change URL from godoc to pkg.go.dev in README.md as godoc is not supported anymore.


manicminer (approved):
Thanks @mkevac, this LGTM 👍

PR #158, opened 2022-03-15
Redacted URL in logs / errors

dany74q:
So that basic auth creds, if any, won't show up in logs


.circleci/config.yml
```diff
@@ -51,5 +51,5 @@ workflows:
       - run-tests:
           matrix:
             parameters:
-              go-version: ["1.14.2"]
+              go-version: ["1.15.15"]
```
dany74q, line 54: Yeah... Apparently the url#Redacted() method was introduced in go 1.15; so it makes this a bigger change than I'd anticipated.

I guess I can just copy go's implementation aside if bumping the go version is unacceptable (understandably, of course)

dany74q: Reverted it and just copied the implementation, no need to potentially break existing users

manicminer (approved):
Many thanks for this fix @dany74q, this LGTM 👍

eastebry:
Hey @dany74q - we've decided to retroactively issue a CVE for this issue. If you would like credit for the discovery or fix, please email security[at]hashicorp.com with the name/identifier you would like us to use (Github user dany74q, Dany from _SomeCompany_, etc.) An acknowledgement will be included in a security bulletin published to https://discuss.hashicorp.com/c/security/

Thank you again!

PR #161, opened 2022-04-11
Retry extra handling

gavriel-hc:
Sometimes a request can "succeed" but fail later and in order to retry the later things, we have to retry the whole request. This PR adds an option to pass a "response handler" to deal with this situation.

The tests were failing (some genuinely, some because of new vet rules)

[Asana](https://app.asana.com/0/1200464981665630/1202082073961730/f)

### DIFF
--- a/client.go
+++ b/client.go
@@ -79,10 +79,10 @@ var (
 // ReaderFunc is the type of function that can be given natively to NewRequest
 type ReaderFunc func() (io.Reader, error)
 
-// ResponseHandlingFunc is a type of function that takes in a Response, and does something with it.
+// ResponseHandlerFunc is a type of function that takes in a Response, and does something with it.
 // It only runs if the initial part of the request was successful.
 // If an error is returned, the client's retry policy will be used to determine whether to retry the whole request.
-type ResponseHandlingFunc func(*http.Response) error
+type ResponseHandlerFunc func(*http.Response) error
 
 // LenReader is an interface implemented by many in-memory io.Reader's. Used
 // for automatically sending the right Content-Length header when possible.
@@ -96,7 +96,7 @@ type Request struct {
 	// used to rewind the request data in between retries.
 	body ReaderFunc
 
-	responseHandler ResponseHandlingFunc
+	responseHandler ResponseHandlerFunc
 
 	// Embed an HTTP request directly. This makes a *Request act exactly
 	// like an *http.Request so that all meta methods are supported.
@@ -114,7 +114,7 @@ func (r *Request) WithContext(ctx context.Context) *Request {
 }
 
 // SetResponseHandler allows setting the response handler.
-func (r *Request) SetResponseHandler(fn ResponseHandlingFunc) {
+func (r *Request) SetResponseHandler(fn ResponseHandlerFunc) {
 	r.responseHandler = fn
 }
 
@@ -589,6 +589,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 	var doErr, respErr, checkErr error
 
 	for i := 0; ; i++ {
+		doErr, respErr = nil, nil
 		attempt++
 
 		// Always rewind the request body when non-nil.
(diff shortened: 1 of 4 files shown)


roundtripper_test.go
```diff
@@ -107,12 +107,12 @@ func TestRoundTripper_TransportFailureErrorHandling(t *testing.T) {
 
 	expectedError := &url.Error{
 		Op:  "Get",
-		URL: "http://this-url-does-not-exist-ed2fb.com/",
+		URL: "http://asdfsa.com/",
```
gavriel-hc, line 110: I don't really understand why, but this was returning a 200 for most URLs, but for whatever reason did not when I typed some gibberish. It seemed potentially related to the ISP since when I went to `this-url-...` in my browser it redirected to some sort of AT&T search page for possible other pages.

Not sure if there is a better way to test this.

client.go
```diff
@@ -445,6 +450,9 @@ func baseRetryPolicy(resp *http.Response, err error) (bool, error) {
 			}
 
 			// Don't retry if the error was due to TLS cert verification failure.
+			if notTrustedErrorRe.MatchString(v.Error()) {
```
ryanuber, line 453: Did you observe some specific condition where the existing `UnknownAuthorityError` check did not catch this? I'd prefer if we kept unrelated fixes like this in separate PRs.

ryanuber (changes requested):
Not bad! I think we could clean up the interface a tad, and I left a comment about the retry trigger being a bool vs an error that I think we should consider.

client.go
gavriel-hc: Yep, this test was failing. https://github.com/hashicorp/go-retryablehttp/blob/master/client_test.go#L577 (specifically with "expected 1 attempt, got 5")

I should probably also update the test a bit more carefully though, since it's probably no longer an x509 error

client.go
```diff
@@ -545,6 +553,11 @@ func PassthroughErrorHandler(resp *http.Response, err error, _ int) (*http.Respo
 
 // Do wraps calling an HTTP method with retries.
 func (c *Client) Do(req *Request) (*http.Response, error) {
+	return c.DoWithResponseHandler(req, nil)
+}
+
+// DoWithResponseHandler wraps calling an HTTP method plus a response handler with retries.
+func (c *Client) DoWithResponseHandler(req *Request, handler func(*http.Response) (shouldRetry bool)) (*http.Response, error) {
```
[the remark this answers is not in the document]
gavriel-hc: Good call, totally agreed on setting it on the Request type.

Re: bool vs error I do think it's important to not _just_ return an error since some errors are retryable and others are not. But I think what I'll do is require it returns both an error and a bool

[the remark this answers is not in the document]
gavriel-hc: Oh the other difficulty is "handling it all together" is a breaking change since `CheckRetry` is exported. To keep things simple for now I'll avoid modifying its signature.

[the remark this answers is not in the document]
gavriel-hc: Actually the more I think about this I'm getting my head in circles. Which is fine except if the interface for consumers of the library makes _their_ heads also spin. Let's chat briefly about this over zoom (we can do so at our 1:1 today)

client.go
gavriel-hc: Never mind, it still is. The full error (in string format) is
`2022/04/13 13:31:14 [ERR] GET https://127.0.0.1:58511 request failed: Get "https://127.0.0.1:58511": x509: “Acme Co” certificate is not trusted`

ryanuber: Strange! In that case, this is totally fine. I wonder if it would be more comprehensive to just replace our origina UnknownAuthority check with an `/x509:/` regex or something like that. There are [a lot of permanent errors](https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/crypto/x509/verify.go) that appear to fall into that category. Either way - if this unblocks you, I'm fine with adding just the one extra check for now, and revisiting this later if needed.

client.go
```diff
@@ -624,6 +642,13 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 			}
 		}
 
+		// Check if we should continue with retries.
+		shouldRetry, checkErr = c.CheckRetry(req.Context(), resp, doErr)
+		if !shouldRetry && doErr == nil && req.responseHandler != nil {
+			respErr = req.responseHandler(resp)
```
ryanuber, line 648: We should log when this returns a non-nil error. Otherwise it wouldn't be clear as to why the client was going to retry. You can see just above here, there's a similar log for when `doErr` is non-nil.

ryanuber (changes requested):
Looks great! Only minor tweaks regarding naming and logging, otherwise this seems solid. 👍

client.go
gavriel-hc: I'll just leave it, and we can revisit later if we ever need to.

ryanuber (approved):
LGTM! 🚢

PR #163, opened 2022-04-25
types for distinguishing retryable/nonretryable errors

gavriel-hc:
Overlooked in previous responseHandler updates, we need a way to distinguish between different categories of errors.

I have mixed feelings as to whether we should handle the unspecified error type as retryable, but it is the simplest with the current way the code is structured (since we are reusing the same retry policy for errors from the request and errors from the response handling, and since we have this [error swallowing](https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L436) built into the default policy)

### DIFF
--- a/client.go
+++ b/client.go
@@ -25,6 +25,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


gavriel-hc:
I still think something like this could be worth thinking about for v2, but isn't necessary at the moment

PR #168, opened 2022-06-23
Properly parsing request body when `http.NoBody` is used

tjasko:
This fix corrects a behavior when [`http.NoBody`](https://pkg.go.dev/net/http#NoBody) is used as the request body, as currently, the `retryablehttp.getBodyReaderAndContentLength()` function will return a `ReaderFunc` that when called returns an `io.Reader`, that will output a literal `"0"` string on the wire as the HTTP request body.

As such, a fix & test has been implemented to no-op when `http.NoBody` is used.

This issue was found as I was testing this library when a `POST` request was being done with no request body.


tjasko:
Rebased against main given how stale this PR is due to lack of review.

PR #177, opened 2022-12-24
Update error log

annawinkler:
Change `ERR` to `ERROR` for consistency in our libraries.

To Do:
* Find out context for the different loggers. When running a `terraform apply` I saw the following log output from this library:

```
2022-12-23T18:19:35.431-0700 [INFO]  CLI command args: []string{"apply"}
2022/12/23 18:19:35 [DEBUG] GET https://app.terraform.io/.well-known/terraform.json
```

The `DEBUG` prints even if `TF_LOG` is set to `info` (which I did in this case). Also would be nice to have consistent date/time stamps.

### DIFF
--- a/client.go
+++ b/client.go
@@ -643,7 +643,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 			case LeveledLogger:
 				v.Error("request failed", "error", err, "method", req.Method, "url", req.URL)
 			case Logger:
-				v.Printf("[ERR] %s %s request failed: %v", req.Method, req.URL, err)
+				v.Printf("[ERROR] %s %s request failed: %v", req.Method, req.URL, err)
 			}
 		} else {
 			// Call this here to maintain the behavior of logging all requests,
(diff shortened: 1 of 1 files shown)


ryanuber (changes requested):
This seems fine on the surface, but what gives me pause is that the `[ERROR]` tag is just what go-hclog has standardized on. It isn't indicative of what every other importer may have standardized on, and I would not be surprised to find if we had users watching for the `[ERR]` tag, specifically. Had you considered changing the caller side to use `go-hclog` if we want such consistency? I think that would be the more correct approach to achieving uniformity in the logging format. Dropping an ❌ while we discuss the impact.

ryanuber:
Regarding the DEBUG logs above, I believe that could also be fixed by use of a leveled logger. Using a non-leveled logger as in the code change here leaves any level filtering to the caller. I think it'd be worth exploring that option to see if we can fix both cases by passing in a structured logger.

annawinkler:
Thanks - I'll explore other options & close this PR.

PR #183, opened 2023-04-03
Add more nil Logger tests

dolmen:
Add more tests with variations of a `nil` Logger beyond just `any(nil)`: `Logger(nil)`, `LeveledLogger(nil)`.


jefferai:
Thanks!

PR #197, opened 2023-05-19
Avoid reading all from bytes.Reader when get request body

longshine:
Avoid reading all bytes from `bytes.Reader` to save memory.

See also `Request.NewRequestWithContext()` in `net/http`.

Ref: https://github.com/golang/go/blob/e71b61180aa19a60c23b3b7e3f6586726ebe4fd1/src/net/http/request.go#L881-L887


manicminer (approved):
Thanks for this improvement @longshine, this LGTM 👍

PR #205, opened 2023-09-08
client.baseRetryPolicy check if response is nil before accessing attributes

tty2:
`baseRetryPolicy`
if accessable from publis `DefaultRetryPolicy` and `ErrorPropagatedRetryPolicy` and it is called without any checks.
There can be case when use call this function with passing nil response inside and catch `panic`.
With this changes they get error with explanation.

### DIFF
--- a/client.go
+++ b/client.go
@@ -28,6 +28,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -492,6 +493,11 @@ func baseRetryPolicy(resp *http.Response, err error) (bool, error) {
 		return true, nil
 	}
 
+	// check if is nil to make below attributes access safe
+	if resp == nil {
+		return false, errors.New("response is nil")
```
manicminer, line 498:
```suggestion
		return true, nil
```

client.go
```diff
@@ -492,6 +493,11 @@ func baseRetryPolicy(resp *http.Response, err error) (bool, error) {
 		return true, nil
 	}
 
+	// check if is nil to make below attributes access safe
```
manicminer, line 496:
```suggestion
	// Retry when the response is nil as this can indicate a closed connection or some other temporary condition
```

manicminer (changes requested):
@tty2 Thanks for submitting this PR. I agree we should be nil-checking the response, however I'd suggest that the request _should_ be retried in that circumstance, since a nil response is often attributed to the connection being closed or the server misbehaving - both temporary conditions that IMO we should smooth over.

PR #207, opened 2023-11-06
Sets request's GetBody field on create wrapper

### LINKED ISSUE #121: Request wrapper make standard request `GetBody` property always nil

sebasslash:
Closes #121 

This commit addresses two key issues:
- Upon receiving a temporary redirect, net/http will only preserve the request's body if GetBody() is [defined](https://pkg.go.dev/net/http#Client.Do) (third to last sentence in the paragraph).
- Upon receiving a GOAWAY frame, the client will create a new connection. We define GetBody() in order to reuse the body sent in the last stream on the now terminated connection.

Not entirely sure why the fix was never merged, but here we are :) Credits to @rgb-24bit

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -978,3 +978,62 @@ func TestClient_StandardClient(t *testing.T) {
 		t.Fatalf("expected %v, got %v", client, v)
 	}
 }
+
+func TestClient_RedirectWithBody(t *testing.T) {
+	var redirects int32
+	// Mock server which always responds 200.
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.RequestURI {
+		case "/redirect":
+			w.Header().Set("Location", "/target")
+			w.WriteHeader(http.StatusTemporaryRedirect)
+		case "/target":
+			atomic.AddInt32(&redirects, 1)
+			w.WriteHeader(http.StatusCreated)
+		default:
+			t.Fatalf("bad uri: %s", r.RequestURI)
+		}
+	}))
+	defer ts.Close()
+
+	client := NewClient()
+	client.RequestLogHook = func(logger Logger, req *http.Request, retryNumber int) {
+		if _, err := req.GetBody(); err != nil {
+			t.Fatalf("unexpected error with GetBody: %v", err)
+		}
+	}
+	// create a request with a body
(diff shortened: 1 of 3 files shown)


dekimsey (approved):
I was wondering if we needed to test for 307/308 response codes before setting the GetBody, but it looks like `Client.Do` is pretty clear that it'll only invoke this helper if it thinks it should. So that works for me!

I 💯 appreciate the additional tests! Thank you

client_test.go
```diff
@@ -978,3 +978,62 @@ func TestClient_StandardClient(t *testing.T) {
 		t.Fatalf("expected %v, got %v", client, v)
 	}
 }
+
+func TestClient_RedirectWithBody(t *testing.T) {
+	var redirects int32
+	// Mock server which always responds 200.
```
brandonc, line 984: Not technically true

client_test.go
```diff
@@ -978,3 +978,62 @@ func TestClient_StandardClient(t *testing.T) {
...
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.RequestURI {
+		case "/redirect":
+			w.Header().Set("Location", "/target")
+			w.WriteHeader(http.StatusTemporaryRedirect)
+		case "/target":
+			atomic.AddInt32(&redirects, 1)
```
brandonc, line 991: Should this test that the body was received?

brandonc (commented):
I am a little late on the review but thought I would share my findings which were not blocking anyway

PR #210, opened 2023-12-01
Fix default retry policy for certificate verification errors and bad request headers

### LINKED ISSUE #176: Go 1.20: fix up retry certificate retry  after Go CL "crypto/tls: add CertificateVerificationError to tls handshake"
Certificate retry inhibiting would be broken in Go 1.20.

The upstream CL http://go.dev/cl/449336 changed the error type returned when there's a certificate verification failure. The following line no longer hits:

https://github.com/hashicorp/go-retryablehttp/blob/493aa4cf372e47ec00228558d9d91a9517b045a5/client.go#L476-L478

I believe it would resume working (and also work for older Go versions) if this was changed to:

```diff
			// Don't retry if the error was due to TLS cert verification failure.
-			if _, ok := v.Err.(x509.UnknownAuthorityError); ok {
+			if _, ok := errors.Is(v.Err, x509.UnknownAuthorityError); ok {
				return false, nil
			}
```

tomclegg:
Go 1.20 started returning `*tls.CertificateVerificationError` instead of `x509.UnknownAuthorityError` for the tested certificate problem. The new error type check doesn't compile in old versions of Go because the type didn't exist yet, so this PR puts the old and new type checks in separate files with appropriate build tags. Fixes #176

This also updates the default policy to disable retry after a "bad (outgoing) header" error from net/http.


manicminer (approved):
Many thanks @tomclegg for this fix, this LGTM 👍

PR #216, opened 2024-03-08
Re-sign request on retry

mgwoj:
In our case we need to re-sign request before initiating retry operation. To achieve that we have added a new handler `PrepareRetry` which can be used for this purpose, however the signature and the name is as generic as possible.

### DIFF
--- a/client.go
+++ b/client.go
@@ -441,7 +441,6 @@ func NewClient() *Client {
 		RetryMax:     defaultRetryMax,
 		CheckRetry:   DefaultRetryPolicy,
 		Backoff:      DefaultBackoff,
-		PrepareRetry: DefaultPrepareRetry,
 	}
 }
 
@@ -556,12 +555,6 @@ func DefaultBackoff(min, max time.Duration, attemptNum int, resp *http.Response)
 		sleep = max
 	}
 	return sleep
-}
-
-// DefaultPrepareRetry is performing noop during prepare retry
-func DefaultPrepareRetry(_ *http.Request) error {
-	// noop
-	return nil
 }
 
 // LinearJitterBackoff provides a callback for Client.Backoff which will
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -435,6 +441,7 @@ func NewClient() *Client {
 		RetryMax:     defaultRetryMax,
 		CheckRetry:   DefaultRetryPolicy,
 		Backoff:      DefaultBackoff,
+		PrepareRetry: DefaultPrepareRetry,
```
manicminer, line 444: Since this default function does nothing, can we leave the default as `nil` here?

client.go
```diff
@@ -551,6 +558,12 @@ func DefaultBackoff(min, max time.Duration, attemptNum int, resp *http.Response)
 	return sleep
 }
 
+// DefaultPrepareRetry is performing noop during prepare retry
+func DefaultPrepareRetry(_ *http.Request) error {
+	// noop
+	return nil
+}
+
```
manicminer, line 566: Not needed if we accommodate a nil value for `client.PrepareRetry`

client.go
```diff
@@ -728,17 +741,24 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		// without racing against the closeBody call in persistConn.writeLoop.
 		httpreq := *req.Request
 		req.Request = &httpreq
+
+		if err := c.PrepareRetry(req.Request); err != nil {
+			prepareErr = err
+			break
+		}
```
manicminer, line 748: Per the above, can we conditionally attempt this if the function is non-nil?

```suggestion
		if c.PrepareRetry != nil {
			if err := c.PrepareRetry(req.Request); err != nil {
				prepareErr = err
				break
			}
		}
```

manicminer (changes requested):
Thank you @mgwoj for this PR! This seems like a reasonable addition and I like the implementation. My only suggestion would be to accommodate a default `nil` value for `Client.PrepareRetry` - this would offer maximum compatibility for consumers who might be constructing their own `Client`.

mgwoj:
Thank you for reviewing and accepting the proposed solution for prepare retry. I have applied your suggestions.

manicminer (approved):
Thanks @mgwoj, this LGTM! 👍

manicminer (approved):

PR #218, opened 2024-05-08
Add maintainer team as code owner

manicminer:


claire-labry (commented):
I think we can remove @hashicorp/release-engineering as we were added as temporary maintainers when we initially migrated this project onto GHA.

manicminer:
@claire-labry Thanks for the review. Sounds good, I've updated that 👍

claire-labry (approved):

Issue #182, opened 2023-02-21
Add support for retrieving logger from `context.Context`


wjam:
It's a common pattern to have a logger saved into the `context.Context` so that various pieces of information can be added to it - trace ID, user ID, etc. The problem is that the library doesn't support passing the context into the logger.

fantapop:
This is also something I'm interested in. There is a similar request against the stripe-go logger with some implementation ideas here: https://github.com/stripe/stripe-go/issues/1281

kayvonkhosrowpour:
We are also very interested in this. I created a zaplogger interface wrapper to meet the LeveledLogger interface that the HTTP client expects (see below). But unfortunately, I can't pass in the context that contains a lot of useful information (X-Correlation-Id, for example).

```golang
// ZapLeveledLogger wrapper for zap.logger for use with hashicorp's go-retryable LeveledLogger
type ZapLeveledLogger struct {
	Logger *zap.Logger
}

// New creates a ZapLeveledLogger with a zap.logger that satisfies standard library log.Logger interface.
func New(logger *zap.Logger) retryablehttp.LeveledLogger {
	return &ZapLeveledLogger{Logger: logger}
}

func (l *ZapLeveledLogger) Error(msg string, keysAndValues ...interface{}) {
	l.Logger.Error(msg, keysAndValuesToFields(keysAndValues)...)
}

func (l *ZapLeveledLogger) Info(msg string, keysAndValues ...interface{}) {
	l.Logger.Info(msg, keysAndValuesToFields(keysAndValues)...)
}

func (l *ZapLeveledLogger) Debug(msg string, keysAndValues ...interface{}) {
	l.Logger.Debug(msg, keysAndValuesToFields(keysAndValues)...)
}

func (l *ZapLeveledLogger) Warn(msg string, keysAndValues ...interface{}) {
	l.Logger.Warn(msg, keysAndValuesToFields(keysAndValues)...)
}

func keysAndValuesToFields(keysAndValues []interface{}) []zapcore.Field {
	fields := []zapcore.Field{}

	// we've already failed if we have an odd number of arguments
	failed := len(keysAndValues)%2 != 0

	for i := 0; (i < len(keysAndValues)-1) && !failed; {
		// get the next key safely, to check if it's a string
		key, hasKey := keysAndValues[i].(string)
		// check that there is an associated value
		hasValue := (i + 1) < len(keysAndValues)

		// if there is a key and value, include it
		if hasKey && hasValue {
[...]

PR #231, opened 2024-06-20
feat: Add support for logging with context

whrss9527:
Add support for logging with context, you can pass slog to NewContextLogger to print trace or other log information.

https://github.com/hashicorp/go-retryablehttp/issues/182


client.go
```diff
@@ -364,6 +364,48 @@ func (h hookLogger) Printf(s string, args ...interface{}) {
...
+
+// contextLogger adapts an ContextLogger to Logger for use by the existing hook functions
+// without changing the API.
+type contextLogger struct {
+	logger ContextLogger
+}
+
+// NewContextLogger returns a new contextLogger
+// for example: NewContextLogger(slog.Default())
+func NewContextLogger(logger ContextLogger) ContextLogger {
```
whrss9527, line 385: A constructor is added here to make it easier for users to use. Due to the existing methods in LeveledLogger, ordinary loggers such as slog cannot be directly passed in and will be identified as LeveledLogger. The constructor is used here for packaging, which makes it easier for external users to use ContextLogger.

whrss9527:
I want to make the logging code as concise and brief as possible, while ensuring the API is user-friendly and compatible with mainstream Golang logging frameworks, such as the increasingly common slog. If you have a better implementation, I am open to learning.

Ehco1996 (commented):
lgtm 

ptal @manicminer

whrss9527:
@Ehco1996 Sorry, I clicked by mistake

client.go
```diff
@@ -364,6 +364,48 @@ func (h hookLogger) Printf(s string, args ...interface{}) {
...
+	ErrorContext(ctx context.Context, msg string, keysAndValues ...interface{})
+	InfoContext(ctx context.Context, msg string, keysAndValues ...interface{})
+	DebugContext(ctx context.Context, msg string, keysAndValues ...interface{})
+	WarnContext(ctx context.Context, msg string, keysAndValues ...interface{})
+}
+
+// contextLogger adapts an ContextLogger to Logger for use by the existing hook functions
+// without changing the API.
+type contextLogger struct {
+	logger ContextLogger
```
mitar, line 380: Please make logger public or embedded, so that it is possible to access the original logger in hooks.

client.go
```diff
@@ -364,6 +364,48 @@ func (h hookLogger) Printf(s string, args ...interface{}) {
...
+func (s contextLogger) DebugContext(ctx context.Context, msg string, keysAndValues ...interface{}) {
+	s.logger.DebugContext(ctx, msg, keysAndValues...)
+}
+
+func (s contextLogger) WarnContext(ctx context.Context, msg string, keysAndValues ...interface{}) {
+	s.logger.WarnContext(ctx, msg, keysAndValues...)
+}
+
+func (s contextLogger) Printf(msg string, keysAndValues ...interface{}) {
+	s.logger.InfoContext(context.Background(), msg, keysAndValues...)
```
mitar, line 406: I think `contextLogger` should have `ctx` as a field, which it then passes on here.

client.go
```diff
@@ -689,6 +733,8 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 
 		if c.RequestLogHook != nil {
 			switch v := logger.(type) {
+			case ContextLogger:
+				c.RequestLogHook(contextLogger{v}, req.Request, i)
```
mitar, line 737: I think you should pass `ctx` to `contextLogger`.

client.go
```diff
@@ -725,6 +773,8 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 			if c.ResponseLogHook != nil {
 				// Call the response logger function if provided.
 				switch v := logger.(type) {
+				case ContextLogger:
+					c.ResponseLogHook(nil, resp)
```
mitar, line 777: Why `nil` here?

client.go
```diff
@@ -832,6 +884,8 @@ func (c *Client) drainBody(body io.ReadCloser) {
 	if err != nil {
 		if c.logger() != nil {
 			switch v := c.logger().(type) {
+			case ContextLogger:
+				v.ErrorContext(context.Background(), "error reading response body", "error", err)
```
mitar, line 888: Why `context.Background()` here?

client.go
whrss9527: Provided for external use is 'ContextLogger' interface, and the role of 'contextLogger' struct is the same as that of 'hookLogger'.

client.go
whrss9527: Here is the method used to implement the Logger interface. It does not add the ctx parameter and maintains the original Printf function.

client.go
mitar: `hookLogger` allows you to access wrapped `Logger`. `contextLogger` does not.

client.go
mitar: But I am saying that it should. The whole point of this PR is that one can access logger and data from the context. And here you are passing `context.Background()` which breaks the whole point of this PR.

client.go
whrss9527: In `RequestLogHook`, you accept the `*http.Request` parameter, and you can get the Context from it. If the user overrides `RequestLogHook`, you can get the Context in this way. However, if you put it in `contextLogger`, it cannot be retrieved and converted externally, because it is not external in design

client.go
whrss9527: You are right, I will revise it, thank you

client.go
whrss9527: Unable to get the context in the request here

client.go
mitar: Yes, `ctx` passed to `contextLogger` does not have be public, but it should be passed to `InfoContext` in `Printf`.

client.go
mitar: Then make `drainBody` accept `ctx`. It is private method so you can change it.

client.go
whrss9527: You're right

client.go
whrss9527: OK~

client.go
whrss9527: Thank you very much. It feels much smoother.

mitar: So maybe just embed `ContextLogger`, the same how `hookLogger` is embedding `LeveledLogger`. Then it can be accessible if needed from hooks.

client.go
mitar: I think `NewContextLogger` now needs `ctx` parameter.

client.go
whrss9527: Yes, I made changes

client.go
whrss9527: I remove it, now you only need to pass in `slog` directly to use it. You don't need to use a constructor to wrap it. If you use another logger externally, you can flexibly wrap it.
What do you think?

Chator1: Ok

mitar: I think this is fine, yes.

mitar (approved):

reedip (commented):
In line #445 in the new code, the comment states that the Custom Logger can be Logger or Leveled Logger. Maybe that can be changed?
This is just a nit, and mainly because I was trying to go through this Client structure and found it.

reedip (approved):

maximderbin:
could this be merged pls?

whrss9527:
ptal @manicminer

RafaeLeal:
I'm also interested in this. Can this be merged? I'm glad to help if anything is missing.

maradanasai:
I'm also interested in this. Can this be merged ?

mitar:
Please , can we get this merged in?

PR #240, opened 2024-09-17
Fix preparing of POST/PUT requests not taking into account request body

dawiddzhafarov:
Hi, in our case we are using a POST body to generate an auth header. The problem is that for the retried requests, it is not taken into account. With this change the request is prepared before sending it and enabling us to construct correct payload. I also added a couple of test cases.

Let me know what you think.

Thanks, 
Dawid

### DIFF
--- a/client.go
+++ b/client.go
@@ -687,6 +687,14 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 			}
 		}
 
+		// First attempt was already signed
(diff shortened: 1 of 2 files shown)


dawiddzhafarov:
Hello,

I already signed the CLA but it seems like it requests the signing again. However when I enter the form, it is already filled up and I cannot sign it again. Could you please verify?

Thanks!
Dawid

PR #251, opened 2025-02-25
Build test

KaushikiAnand:
- I have added unit test coverage and linting to the ci.yml workflow. For more background, refer the https://hashicorp.atlassian.net/browse/IND-2338

- The unit test coverage change was done to help understand the defects early in lifecycle and come up with a good solution for it. The linting was done to ensure that the codebase is consistent and maintainable and helps in improving the code quality and reducing errors.


.github/workflows/pr-unit-tests-1.20.yaml
```diff
@@ -7,11 +7,28 @@ jobs:
...
     steps:
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Checkout Code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
 
-      - uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
+      - name: Setup Go
+        uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
         with:
           go-version: 1.22
```
sonamtenzin6, line 16: Suggestion: You can change the name of the file to *-1.22.yaml

client.go
```diff
@@ -670,7 +670,7 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 	var doErr, respErr, checkErr, prepareErr error
 
 	for i := 0; ; i++ {
-		doErr, respErr, prepareErr = nil, nil, nil
+		//doErr, respErr, prepareErr = nil, nil, nil
```
sonamtenzin6, line 673: Question: Why did we need to comment out this statement ?

client_test.go
```diff
@@ -223,7 +223,14 @@ func testClientDo(t *testing.T, body interface{}) {
...
 	defer list.Close()
-	go http.Serve(list, handler)
+	errors := make(chan error)
+	go func() {
+		err := http.Serve(list, handler)
+		if err != nil {
+			errors <- err
+			return
+		}
+	}()
```
sonamtenzin6, line 233: Question: Was this added as a fix for lint check ?

Also, we should check if there were any errors added to `errors` and see if the test needs to fail

client.go
KaushikiAnand: this was coming as an unused linter issue

KaushikiAnand: I commented out this statement as a  part of unused linter issue.
Should I disable it like I did in previous prs?

client_test.go
KaushikiAnand: Yes this was added as a part of lint fix.

KaushikiAnand: go func() {
		for err := range errors {
			log.Println("Received error:", err)
		}
}()
on adding this I saw that error. was being passed to the errors channel, and the test is designed to fail.

sonamtenzin2 (approved):

PR #259, opened 2025-05-31
add comment for bodyType param

Yiming1997:
Add comment for bodyType param to avoid confusion during usage

### DIFF
--- a/client.go
+++ b/client.go
@@ -871,26 +871,12 @@ func (c *Client) Head(url string) (*http.Response, error) {
 
 // Post is a shortcut for doing a POST request without making a new client.
 // The bodyType parameter sets the "Content-Type" header of the request.
-// Common content type values include:
-//   - "application/json"          : For JSON-encoded data
-//   - "application/xml"           : For XML payloads
-//   - "text/plain"                : For plain text
-//   - "application/octet-stream"  : For binary data
-//   - "application/x-www-form-urlencoded": For form submission data
-//   - "multipart/form-data"       : For file uploads and complex forms
 func Post(url, bodyType string, body interface{}) (*http.Response, error) {
 	return defaultClient.Post(url, bodyType, body)
 }
 
 // Post is a convenience method for doing simple POST requests.
 // The bodyType parameter sets the "Content-Type" header of the request.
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -870,11 +870,27 @@ func (c *Client) Head(url string) (*http.Response, error) {
...
 
 // Post is a shortcut for doing a POST request without making a new client.
+// The bodyType parameter sets the "Content-Type" header of the request.
+// Common content type values include:
+//   - "application/json"          : For JSON-encoded data
+//   - "application/xml"           : For XML payloads
+//   - "text/plain"                : For plain text
+//   - "application/octet-stream"  : For binary data
+//   - "application/x-www-form-urlencoded": For form submission data
+//   - "multipart/form-data"       : For file uploads and complex forms
```
abhijeetviswa, line 880: Are examples required here? `Content-Type` is well documented in the spec.

client.go
```diff
@@ -870,11 +870,27 @@ func (c *Client) Head(url string) (*http.Response, error) {
...
 
 // Post is a convenience method for doing simple POST requests.
+// The bodyType parameter sets the "Content-Type" header of the request.
+// Common content type values include:
+//   - "application/json"          : For JSON-encoded data
+//   - "application/xml"           : For XML payloads
+//   - "text/plain"                : For plain text
+//   - "application/octet-stream"  : For binary data
+//   - "application/x-www-form-urlencoded": For form submission data
+//   - "multipart/form-data"       : For file uploads and complex forms
```
abhijeetviswa, line 893: Same as above.

abhijeetviswa (changes requested):

client.go
Yiming1997: thanks for your review! I think we can remove these examples.

But can we leave the descriptions for bodyType? Because for some users it could cause confusions when seeing it at the first time. when user use it ,they might not realize the param bodyType actually is for setting content -type

client.go
Yiming1997: ok ,already removed,thank u!

abhijeetviswa (approved):

PR #282, opened 2026-04-13
client: add RedactQueryParams to hide query string from logs and errors

### LINKED ISSUE #206: client logs query parameters, which can leak secrets
There are several places where `url.URL` is used in a log or error, for example:

via logs:
https://github.com/hashicorp/go-retryablehttp/blob/571a88bc9c3b7c64575f0e9b0f646af1510f2c76/client.go#L597

via returned error:
https://github.com/hashicorp/go-retryablehttp/blob/571a88bc9c3b7c64575f0e9b0f646af1510f2c76/client.go#L751

While the former can be mitigated by setting a `nil` logger on client instantiation, users of the client have to deal with potential query param / data leaks through errors.

We would like to see either a removal of query params entirely, or an option to disable full query param reporting. We could replace instances of `url.URL` with a few options:

```
# Option 1, build URL w/o query params
cleanURL := url.Scheme + "://" + url.Host + url.Path

# Option 2, truncate query params
cleanURL := strings.Split(url.URL, "?")[0]
```

alliasgher:
## Summary

Query parameters frequently carry credentials — bearer tokens, presigned-URL signatures, OAuth tokens — yet today every request URL (query string included) is copied verbatim into debug logs **and** into the returned \`giving up after N attempt(s)\` errors. Callers that can't swap the logger still leak the query through errors.

This PR adds an opt-in \`Client.RedactQueryParams bool\` field. When set to \`true\`, the query string and fragment are stripped before the URL is rendered in logs and errors. The default is \`false\`, so existing behavior is preserved. Userinfo passwords continue to be redacted to \`xxxxx\` regardless of the setting.

Internally:

- \`redactURL\` now delegates to a small parameterized \`redactURLFor\` helper so the stripping policy flows through without changing any exported surface besides the new field.
- All call sites in \`Client.Do\` go through a new \`Client.loggableURL\` helper that consults \`RedactQueryParams\`.

### Example

\`\`\`go
c := retryablehttp.NewClient()
c.RedactQueryParams = true

// If the call fails after N retries, the returned error reads:
//   GET https://api.example.com/resource giving up after 3 attempt(s): ...
// instead of:
//   GET https://api.example.com/resource?signature=AKIA...&token=hunter2 giving up after 3 attempt(s): ...
\`\`\`

Closes #206

## Testing

- Added \`TestRedactURLFor\` (6 sub-cases) covering password redaction, query stripping, fragment stripping, nil-safe path.
- Added \`TestClient_RedactQueryParams_InError\` asserting that the query secret does not appear in the returned error after retries are exhausted.
- Full \`go test ./...\` suite passes.

### DIFF
--- a/client.go
+++ b/client.go
@@ -432,6 +432,13 @@ type Client struct {
 	// PrepareRetry can prepare the request for retry operation, for example re-sign it
 	PrepareRetry PrepareRetry
 
+	// RedactQueryParams, when true, strips the query string from request URLs
+	// before they appear in logs and returned errors. This is useful because
+	// query parameters often carry credentials (tokens, signatures, etc.) and
+	// leaking them into error messages or log output can be a security hazard.
+	// Userinfo passwords are always redacted regardless of this setting.
+	RedactQueryParams bool
+
 	loggerInit sync.Once
 	clientInit sync.Once
 }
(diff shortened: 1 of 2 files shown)


orf:
Without this change it's really difficult to reliably redact URLs. Our specific case involves retrieving data using S3 signed urls, and it is very important that these are not inadvertently logged.

PR #283, opened 2026-05-09
fix: cap Retry-After sleep to RetryWaitMax in DefaultBackoff

### LINKED ISSUE #247: Default backoff doesn't honour RetryWaitMax when "Retry-After" header is sent in the response
https://github.com/hashicorp/go-retryablehttp/blob/9dfd949154e507c91ee569367ffa461991b80950/client.go#L551-L566

The DefaultBackoff in function has a logical flaw when handling the `Retry-After` header. If the server provides an unreasonably high value in the Retry-After header, the function respects it without enforcing the `RetryWaitMax` limit set for the httpClient. This can lead to indefinite wait times or blocking behaviour. 

### Steps to Reproduce
1. Simulate a server response with an HTTP `429 Too Many Requests` status.
2. Set a high value in the `Retry-After` header (e.g., `Retry-After: 3600` for 1 hour).
3. Observe that the backoff duration exceeds the configured `RetryWaitMax` limit.

### Expected Behaviour
I believe the backoff duration should always be bounded by the configured `RetryWaitMax` value, regardless of the `Retry-After` header value.

### Suggested Fix
Update the function to enforce the `max` limit after parsing the `Retry-After` header:
```go
if sleep > max {
    return max // Enforce max backoff limit
}
```

nghiack7:
## Problem

When a server responds with a `Retry-After` header (e.g. `Retry-After: 3600`), `DefaultBackoff` unconditionally sleeps for that duration — completely ignoring the client's configured `RetryWaitMax`. This means a misbehaving or malicious server can force a client to sleep for an arbitrarily long time, bypassing the user's intent.

Fixes #247.

## Root cause

```go
// before — no cap applied
if sleep, ok := parseRetryAfterHeader(resp.Header["Retry-After"]); ok {
    return sleep   // could be hours
}
```

## Fix

Cap the `Retry-After` sleep at `max` (which is `RetryWaitMax`):

```go
if sleep, ok := parseRetryAfterHeader(resp.Header["Retry-After"]); ok {
    if sleep > max {
        return max
    }
    return sleep
}
```

This is consistent with how the exponential backoff path already respects `RetryWaitMax`.

## Tests

Added `TestClient_DefaultBackoff_RetryAfterExceedsMax` covering:
- HTTP 429 with `Retry-After: 3600` → capped to `RetryWaitMax` (10s)
- HTTP 503 with `Retry-After: 3600` → capped to `RetryWaitMax` (10s)

All existing tests pass.

### DIFF
--- a/client.go
+++ b/client.go
@@ -552,6 +552,9 @@ func DefaultBackoff(min, max time.Duration, attemptNum int, resp *http.Response)
 	if resp != nil {
 		if resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode == http.StatusServiceUnavailable {
 			if sleep, ok := parseRetryAfterHeader(resp.Header["Retry-After"]); ok {
+				if sleep > max {
+					return max
+				}
 				return sleep
 			}
 		}
--- a/client_test.go
+++ b/client_test.go
@@ -896,6 +896,51 @@ func TestClient_DefaultBackoff(t *testing.T) {
 	}
 }
 
+// TestClient_DefaultBackoff_RetryAfterExceedsMax verifies that DefaultBackoff
+// caps the sleep duration from a Retry-After header at RetryWaitMax so that
+// a server cannot force an arbitrarily long wait.
+func TestClient_DefaultBackoff_RetryAfterExceedsMax(t *testing.T) {
+	testStaticTime(t)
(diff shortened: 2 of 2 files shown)


nghiack7:
@hashicorp-cla-app recheck

PR #288, opened 2026-06-30
fix: preserve final response in PassthroughErrorHandler

### LINKED ISSUE #156: PassthroughErrorHandler shouldn't pass an error if response is non-nil
The README for this package reads:

> [...] if an error is returned by the client (connection errors, etc.), or if a 500-range response code is received (except 501), then a retry is invoked after a wait period. Otherwise, the response is returned and left to the caller to interpret.

However, as it stands, `PassthroughErrorHandler` doesn't seem to behave in a way that would accomplish this behaviour out of the box under normal circumstances. This is because it always passes to the caller of `RoundTrip` the final `http.Response` and the error from the call to `CheckRetry`. A default `http.Client` [will actually throw away the response](https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/net/http/client.go;l=255;drc=refs%2Ftags%2Fgo1.17.8) if an error is returned. So in the case where, for example, we exhaust all `RetryMax` attempts due to consecutive 5xx errors, the caller will never see the final response body. 

`PassthroughErrorHandler` could be much more useful if it first checks for a non-`nil` response:

```go
func PassthroughErrorHandler(resp *http.Response, err error, _ int) (*http.Response, error) {
	if resp != nil {
		return resp, nil
	}
	return resp, err
}
```

This throws away the helpful `CheckRetry` error on the last request, but gives the caller the expected ability to distinguish unexpected HTTP status codes from connection errors. Or maybe I'm misinterpreting what purpose `PassthroughErrorHandler` has?

happysnaker:
## Description

- make `PassthroughErrorHandler` return the terminal `*http.Response` without a competing error when a response exists
- keep `http.Client` / `StandardClient()` callers from silently discarding the final HTTP response on retry exhaustion
- add focused tests for the error handler itself and the `StandardClient` / `RoundTripper` path

## Related Issue

Closes #156

## How Has This Been Tested?

- added unit coverage for `PassthroughErrorHandler`
- added an integration-style test covering `StandardClient()` preserving the final 500 response

### DIFF
--- a/client.go
+++ b/client.go
@@ -656,9 +656,13 @@ func RateLimitLinearJitterBackoff(min, max time.Duration, attemptNum int, resp *
 }
 
 // PassthroughErrorHandler is an ErrorHandler that directly passes through the
-// values from the net/http library for the final request. The body is not
-// closed.
+// final response when one exists. This keeps standard-library callers such as
+// http.Client from discarding the response because of a non-nil error. The
+// body is not closed.
 func PassthroughErrorHandler(resp *http.Response, err error, _ int) (*http.Response, error) {
+	if resp != nil {
+		return resp, nil
+	}
 	return resp, err
 }
 
--- a/client_test.go
+++ b/client_test.go
@@ -626,6 +626,21 @@ func testClientRequestLogHook(t *testing.T, logger interface{}) {
 	}
 }
 
+func TestPassthroughErrorHandler_ReturnsResponseWithoutError(t *testing.T) {
+	resp := &http.Response{
+		StatusCode: http.StatusInternalServerError,
+		Body:       io.NopCloser(strings.NewReader("retry me")),
+	}
+
+	gotResp, gotErr := PassthroughErrorHandler(resp, errors.New("retry stopped"), 3)
+	if gotResp != resp {
+		t.Fatalf("expected same response pointer to be returned")
+	}
+	if gotErr != nil {
+		t.Fatalf("expected nil error when response is present, got %v", gotErr)
+	}
+}
+
 func TestClient_ResponseLogHook(t *testing.T) {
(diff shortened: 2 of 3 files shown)


happysnaker:
This should now be ready for review on my side. The current status check blocker is the HashiCorp CLA gate; I have already signed the CLA before, but the new PR still shows `license/cla` as pending right after creation, so I am waiting for that bot/status to refresh.

Change summary:
- `PassthroughErrorHandler` now returns the terminal `*http.Response` without a competing error when a response exists
- added a focused unit test for the handler
- added a `StandardClient()` / `RoundTripper` test showing the final 500 response body remains available to callers

If you want, I can also add a short README/doc note clarifying this default behavior for `PassthroughErrorHandler`.

happysnaker:
Follow-up: I force-pushed the branch so the commit is now authored with my GitHub-linked noreply address (`73147033+happysnaker@users.noreply.github.com`) instead of the previous unlinked work email.

If the CLA check does not refresh automatically from that updated commit metadata, please let me know and I can re-run whatever recheck flow you prefer.

happysnaker:
Quick follow-up on the remaining blocker: the commits on these branches now use my GitHub-linked noreply address, and I've already completed the HashiCorp CLA flow, but `license/cla` is still stuck in `PENDING` on #288 / #289 / #290.

If there is a maintainer-side recheck I should trigger, or a preferred support path for a stuck CLA status, I'd appreciate a pointer. Thanks!

happysnaker:
Update: I re-ran the HashiCorp CLA flow successfully and the CLA backend now reports my signature as active for this repo/PR context. If the GitHub `license/cla` status remains stuck in `PENDING`, it looks like a sync/recheck issue rather than an unsigned CLA on my side.

PR #289, opened 2026-06-30
fix: avoid panics on typed nil request bodies

### LINKED ISSUE #188: panic on nil *bytes.Buffer as body
When `NewRequestWithContext` is given a `nil` *bytes.Buffer as the body, it panics. Best guess is that it is because the type of the body is being checked, but not that it is non-nil.

Reproduction code ([example here](https://go.dev/play/p/6AYSNl6ErJB)):

```go
package main

import (
	"bytes"
	"context"
	"net/http"

	"github.com/hashicorp/go-retryablehttp"
)

type config struct {
	Method string
	URL    string
	Body   *bytes.Buffer
}

func main() {
	cfg := config{
		Method: http.MethodGet,
		URL:    "https://example.com",
	}
	_, _ = retryablehttp.NewRequestWithContext(
		context.Background(),
		cfg.Method,
		cfg.URL,
		cfg.Body,
	)
}
```

Panic output:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x5edb1c]

goroutine 1 [running]:
bytes.(*Buffer).Len(...)
	/usr/local/go-faketime/src/bytes/buffer.go:73
github.com/hashicorp/go-retryablehttp.getBodyReaderAndContentLength({0x63e680?, 0x0?})
	/tmp/gopath1053138375/pkg/mod/github.com/hashicorp/go-retryablehttp@v0.7.2/client.go:228 +0x35c
github.com/hashicorp/go-retryablehttp.NewRequestWithContext({0x6afd98, 0xc0000ac000}, {0x648621, 0x3}, {0x64c514, 0x13}, {0x63e680?, 0x0?})
	/tmp/gopath1053138375/pkg/mod/github.com/hashicorp/go-retryablehttp@v0.7.2/client.go:295 +0x5a
main.main()
	/tmp/sandbox1468963018/prog.go:22 +0x7b
```

happysnaker:
## Summary
- treat typed nil pointer bodies as no body instead of dereferencing them
- fix the panic reported for `*bytes.Buffer(nil)`
- add focused regression coverage for typed nil `*bytes.Buffer` and `*bytes.Reader` request bodies

## Why
The current `getBodyReaderAndContentLength` type-switch handles concrete pointer body types like `*bytes.Buffer` and `*bytes.Reader`, but it assumes the pointer value is non-nil once the dynamic type matches. That means a typed nil body can panic during request construction instead of behaving like an absent body.

This patch treats typed nil pointer/function bodies the same way as a nil body and keeps the resulting request behavior aligned with the existing no-body path.

Closes #188.

### DIFF
--- a/client.go
+++ b/client.go
@@ -35,6 +35,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
@@ -205,6 +206,10 @@ func getBodyReaderAndContentLength(rawBody interface{}) (ReaderFunc, int64, erro
 	var bodyReader ReaderFunc
 	var contentLength int64
 
+	if isNilRawBody(rawBody) {
+		return nil, 0, nil
+	}
+
 	switch body := rawBody.(type) {
 	// If they gave us a function already, great! Use it.
 	case ReaderFunc:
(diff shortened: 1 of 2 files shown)


happysnaker:
This should be ready for review on my side.

I have already signed the HashiCorp CLA previously, but the `license/cla` check is still showing as pending right after PR creation, so I'm waiting for that status to refresh.

Change summary:
- treat typed nil pointer bodies like `*bytes.Buffer(nil)` as no body instead of dereferencing them
- keep request construction aligned with the existing nil-body path
- add focused regression coverage for typed nil `*bytes.Buffer` and `*bytes.Reader` bodies

PR #290, opened 2026-06-30
fix: avoid panics on unsupported logger types

### LINKED ISSUE #243: Using custom logger type causes panic at runtime
ref: [main/client.go](https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L458-L464)

Can we Update the `retryablehttp.Client.Logger` type struct field to be either

1. a composite interface (ala go 1.18+ generics)?  

```
type RetryableHttpLoggerConstraint interface {
    retryablehttp.Logger | retryablehttp.LeveledLogger
}
func SetLogger[T LoggerConstraint](client *retryablehttp.Client, logger T) {
    client.Logger = logger
}
```

2. a single type and use that vs `interface{}`
3. something more specific than literally anything

Doing a type check and panic-ing at runtime is no bueno and is likely to cause incidents in production (like it did with us).

happysnaker:
## Summary
- stop treating unsupported `Client.Logger` values as a runtime panic
- fall back to no-op internal logging when the logger does not implement `Logger` or `LeveledLogger`
- add regression coverage showing requests and log hooks still complete without crashing

## Why
The current `Client.logger()` path panics during request execution when `Client.Logger` is set to an unsupported type. That turns a configuration mistake into a production crash.

This patch keeps the existing supported logger behavior unchanged, but treats unsupported logger values the same way as no logger for internal logging paths. Requests still complete, hooks still run, and applications no longer crash at runtime because of an invalid logger assignment.

Closes #243.

### DIFF
--- a/client.go
+++ b/client.go
@@ -432,7 +432,6 @@ type Client struct {
 	// PrepareRetry can prepare the request for retry operation, for example re-sign it
 	PrepareRetry PrepareRetry
 
-	loggerInit sync.Once
 	clientInit sync.Once
 }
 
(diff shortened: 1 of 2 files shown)


happysnaker:
This should be ready for review on my side.

I have already signed the HashiCorp CLA previously, but the `license/cla` check is still showing as pending right after PR creation, so I'm waiting for that status to refresh.

Change summary:
- stop treating unsupported `Client.Logger` values as a runtime panic
- fall back to no-op internal logging for unsupported logger types
- add regression coverage showing requests and log hooks still complete without crashing

